PR #7, opened 2018-03-31
Change some Agents to Concept

bgyori:
With the latest change in INDRA, we can use the more general `Concept` class as arguments of Influences, with `Agent` meant to be used for molecular agents. Looks like the only place this makes a difference here is in the test which I updated.


adarshp:
Thanks for the changes!

PR #33, opened 2018-07-05
Major refactor

adarshp:
This PR refactors the core module into an assembly, visualization, export, core, and api module. 

The api module is a significant addition - we expect to keep this API as stable and well-documented as possible.

Also, some tools are added to explore CAGs using Jupyter notebooks.


cl4yton:
+1

PR #39, opened 2018-07-13
Fix Pipenv integration with Travis CI

adarshp:
This PR fixes up Pipenv integration with Travis CI that arise due to NetworkX version mismatches between INDRA and Delphi.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,6 @@ python:
 install:
   - export PYTHONPATH=$PYTHONPATH:$TRAVIS_BUILD_DIR
   - pip install -U pipenv
-  - pipenv install -d
-  - pipenv install networkx==2.1 --skip-lock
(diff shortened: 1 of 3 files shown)


cl4yton:
Have you heard from INDRA folks about when they might complete upgrading NetworkX?

cl4yton:
Thanks @bgyori !

adarshp:
Yeah, for now the way to go is to do `pipenv install -d` followed by `pipenv install networkx==2.1` which overrides NetworkX 1.11 which is installed as a dependency of INDRA.

cl4yton:
Go it.  Should we update the <delphi_root>/README.md?  It currently says to do just `pipenv install -d --ignore-pipfile`.  I can replace with the above two step if you concur.

adarshp:
Yes, thank you!

PR #40, opened 2018-07-17
Modifying internal representation of Delphi in preparation for program analysis

adarshp:
This PR modifies the internals of Delphi so that updating a latent state of the DBN is no longer a matrix multiplication, but a collection of local updates - each CAG node now has a configurable update function. 

@pauldhein : In line 515, a 'default' update function is associated with each node. I imagine that this default could be overridden by update functions constructed based on program analysis.


cl4yton:
Thanks Adarsh, this all sounds great and I fully support you having gone ahead and merged.

PR #44, opened 2018-07-23
Using python3 for autoTranslate

pauldhein:
As opposed to using `python3.6` specifically in the autoTranslate script we should use `python3` so that 3.7 can be used to run `autoTranslate.sh`. I have tested this on my own system and was able to run the script with no errors when using `python3.7`. 

@stephensj2 does this change work for your system or do you require the use of `python3.6`?


stephensj2:
The reason I've got python3.6 in there is because of an issue with the machine I use. Orion (the server I work on) has python3 aliased with python 3.5. It is apparently not recommended to switch this to a newer version of python, so I had to install python 3.6 as python3.6. I can adjust the script to use python3 if the version is above a certain version. As Paul pointed out, Python 3.6 is a lower bound.

adarshp:
@stephensj2 : Yes, that would be great if you could make that adjustment. Thanks!

PR #47, opened 2018-07-23
Refactoring genPGM.py

adarshp:
This PR refactors genPGM.py for greater modularity.


adarshp:
@stephensj2 I made some changes to `genPGM.py` - I don't think they should be too controversial - let me know if you approve and I'll merge into `pa_dev`.

PR #48, opened 2018-07-23
Changing import path so it works on Jons server

adarshp:
@stephensj2 I changed it to `from .genCode import *` - does that work for you? If so, please merge this commit as well.

### DIFF
--- a/delphi/program_analysis/autoTranslate/scripts/genPGM.py
+++ b/delphi/program_analysis/autoTranslate/scripts/genPGM.py
@@ -8,7 +8,7 @@ import re
 import argparse
 from functools import *
 import json
-from .genCode import *
+from delphi.program_analysis.autoTranslate.scripts.genCode import *
 from typing import List, Dict
 from itertools import chain
 import operator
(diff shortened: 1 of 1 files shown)


stephensj2:
It still doesn't like it. Here's the error I get now:

Traceback (most recent call last):
  File "scripts/genPGM.py", line 11, in <module>
    from .genCode import *
ModuleNotFoundError: No module named '__main__.genCode'; '__main__' is not a package

adarshp:
@stephensj2 I reverted the last commit - do you have the delphi source directory in your $PYTHONPATH?

stephensj2:
I do not.

adarshp:
Can you add it and see if having 

`from delphi.program_analysis.autoTranslate.scripts.genCode import *` works?

On my computer, I have the following line in my `~/.bash_profile`: 

```bash
export PYTHONPATH=~/ml4ai/delphi/:$PYTHONPATH
```

stephensj2:
Yeah, that works.

PR #63, opened 2018-08-17
Fixing loop indexing

adarshp:


adarshp:
This PR implements proper tracking of the loop index when converting FORTRAN source code to a DBN using program analysis. This is done by using a boolean node attribute called 'visited'.

When the update function is called for the ProgramAnalysisGraph, each node in it is updated if it has not been 'visited' yet. The predecessors of the node are 'visited' and updated recursively as well, ensuring that no node is doubly updated.

Also, the starting value of the loop index variable is now set by extracting the value from the PGM JSON rather than set to 1 by default.

Also, code formatting and bugfixes, including one when no mapping file is provided for the `map_concepts_to_indicators` function.

@pauldhein Please review the changes to `scopes.py` and let me know if they are okay with you. (but don't merge the PR yet).

delphi/program_analysis/ProgramAnalysisGraph.py
```diff
@@ -8,69 +8,88 @@
...
+        G.nodes[name]["is_index"] = True
+        G.nodes[name]["value"] = int(n.attr["start"])
+        G.nodes[name]["visited"] = True
+        G.nodes[name]["update_fn"] = (
+            lambda **kwargs: int(kwargs.pop(list(kwargs.keys())[0])) + 1
+        )
+        G.add_edge(name, name)
+
+
+def add_action_node(A: AGraph, G: nx.DiGraph, λs, n):
```
pauldhein: Not so wild about having unicode characters in our code. This seems like a large decrease in writability to me as it takes time to learn letter locations for different keymaps. It takes much less time for me to type `lambdas` when trying to reference a variable name than it does for me to google how to type the character lambda on a mac (and I am unwilling to just copy-pasta a character).

Also using lambda here gives me the first-impression that this variable is related to the mathematical usage of lambda; however, I believe this particular usage is closer to the colloquial CS usage.

Please see this [StackExchange post](https://softwareengineering.stackexchange.com/questions/16010/is-it-bad-to-use-unicode-characters-in-variable-names) for an extended discussion on this matter.

adarshp: Fair enough (although lambda is descended from the lambda calculus, which was invented by a mathematician :) )

I have no problem removing the unicode character though.

adarshp:
@pauldhein I've removed the Unicode characters - if this looks good to you, please merge this branch into master and delete it.

PR #101, opened 2018-10-31
Multi-model ForwardInfluenceBlanket creation

pauldhein:
This branch is not ready to be merged, but it is ready to be discussed, and this PR will serve as a platform for discussion. (This really aught to be a separate project or set of issues)

## What is Implemented
- A function for getting all of the input nodes of an extracted model network
- A function for getting all of the output nodes for an extracted model network
- A function for getting the set of all simple paths between each input/output node pair (This information is returned as a dictionary where the keys are input/output node pairs and the values are lists of paths between the input/output node

## What Needs to be Implemented
- A function that takes in the simple path dictionaries of two graphs and uses this to create simplified graphs over an equivalent node set. We will then perform pre-defined similarity functions over this node set (perhaps graph edit-distance) to determine the diff between these simplified graphs
- A function that takes in the simple path dictionaries of two graphs and determines which I/O paths in each graph are not present in the other graph and are therefore automatically included in a diff between the two graphs 

## Design Decisions
- What should these extracted model graphs be called? Should they have their own data structure (currently they are DiGraphs)?
- Where should these new functions for operating on extracted model graphs live? In a class or just a module?

## Similarity Metric Questions
- How do we go from a program DBN to an extracted model graph?
- How should we include information about domains of nodes as well as the domains of nodes when doing diffing? ( i.e. What if A --> B --> C -- > D is equivalent to A --> E --> D because domain(E) == domain(B) and range(E) == range(C) and E(x) ==( C º B)(x) would this mean that the two paths are equivalent?)


adarshp:
Thanks @pauldhein !

## Design decisions

- I suggest making them instances of `AnalysisGraph`, which itself subclasses `networkx.DiGraph`. You could then add a `@classmethod` 'constructor'  in `AnalysisGraph.py` to create them from DOT files. I think this would be good for a couple of reasons:
  - The causal analysis graph is meant to be the central data structure for the World Modelers program
   - The end product of the AutoMATES program is supposed to be a semantically rich model that unifies information from text, equations, and software (A hope of mine is to somehow unify the `AnalysisGraph` and `ProgramAnalysisGraph` classes someday)
- I have a preference for having the new functions live in a module from personal experience - I originally had the functions in the `bmi`, `execution`, `export`, `inspection`, `manipulation`, `parameterization`, `quantification`, and `visualization` modules as methods of the `AnalysisGraph` class, but I ended up splitting them out into different modules - I found it easier to design unit and integration tests this way, plus I wanted to avoid creating a God Object. Anyway, those are my thoughts, but YMMV and I'm certainly open to being convinced otherwise. I suggest creating a module named `model_comparison` or something similar to hold these functions.

## Similarity Metric
- I implemented a constructor for the `ProgramAnalysisGraph` class that takes a 'computational' DBN (i.e. a PyGraphviz AGraph object that is created and returns a DiGraph whose link structure encodes the dependency structure between the variables:

https://github.com/ml4ai/delphi/blob/02e31b26bcbc408ad8ed53bed6ebdcaa5c2925f7/delphi/program_analysis/ProgramAnalysisGraph.py#L81-L120

Perhaps this code could be a starting point? (You can see the input and output in cells [23] and [24] of [this notebook](http://vision.cs.arizona.edu/adarsh/Delphi-Demo-Notebook.html).)
- Re: domains/ranges, etc. Are A, B, C, and D functions? If so, does

  A → B → C → D ≡ D o C o B o A?

  If yes, then

  E = C o B ⇒ D o C o B o A ≡ D o E o A

pauldhein:
# PR Update
This PR now includes the `CausalMarkovBlanket` (CMB) class. This class creates a networkx DiGraph from all paths between all shared components of two models. This allows for analysis to be done over the shared components. As the name suggests, all predecessors of any shared component are also included in the CMB. Testing code has been added to ensure that the cover sets of nodes for the ASCE and PT models are accurate when building CMBs from comparing these two models. The graph structure of a CMB can easily be obtained using the `draw_graph()` function as you would on a normal DiGraph. 

@cl4yton we should review what we expect to be the CMB output for the ASCE one more time before merging this PR.

@adarshp I'm not sure why the Travis builds are failing, can you help with that?

pauldhein:
@adarshp hold off on merging this branch. Just discovered that my CMB is not capturing the parents of child nodes of nodes that are members of the reduced graph. I'll update this PR once I have that fixed.

adarshp:
@pauldhein The Travis builds were failing for a couple of reasons:
1. The reason I told you about earlier - I had updated the table of indicator data in the Google Drive Delphi data folder, but not scp-ed the new version of `delphi.db` to vision.
2. The tests were looking for `.dot` files in `data_dir`, but Travis doesn't have access to the Google Drive folder that `data_dir` points to (i.e. we should use `data_dir` only for dev work and scripts that are not to be used by end users - some of the scripts in `delphi/scripts` use `data_dir`, but we don't really mention it in the `README.md` of the repo. At some point when I get around to updating the `CONTRIBUTING.md` file with more detailed instructions for developers, I'll add a note about `data_dir`). The solution to this was to copy over the `.dot` files to `tests/data/pa_graph_examples`. In general, we should keep files (text files, not binary/pickled) that are required for testing but are not too large in `tests/data` and commit them along with the tests.

pauldhein:
## 2019-01-23 Post meeting design changes
The blanket graph class is now named ForwardInfluenceBlanket to correctly capture what we are modeling. We are not capturing the full markov blanket because (currently) we are only interested in doing forward analysis. As such the ForwardAnalysisBlanket does not capture the child nodes that do not lead to the included outputs, nor does it capture said child nodes parents, and this is by design for the reason stated above.

PR #108, opened 2018-11-09
Process CliMIS rainfall data

bgyori:
This PR adds a function to process the rainfall data. It gets values from multiple CSV files with state-level data and one XLS file with county-level data.

NOTE: I renamed the file from 12_month_evaluation.py to month12_evaluation.py because if the script starts with a number it can't be imported from.


adarshp:
This looks great, thanks @bgyori !

PR #110, opened 2018-11-14
Implemented Background Task under Test Mode

JiamingHao:
I just made the background task work using celery under test mode. To see it, need to run the command "pipenv run celery -A api.celery worker" first in icm_api directory and then run the test.


JiamingHao:
I pushed the version with my way to work around the circular dependency issure we discussed on Wednesday. I added the run_test_icm_api.py in the tests directory and the configuration setting in config.py is now for normal running instead of testing. What the run_test_icm_api.py does is that when you invoke this code, it will first overwrite the setting in the config.py with the one used for tests, and it will activate the worker in the background in a subprocess. After that, it will also run the pytests and show the result. When the above steps are done, it will restore the content in the config.py to the original one used for normal running. So, in my solution, to run the pytests and see the result, the command is "pipenv run python run_test_icm_api.py", the run_test_icm_api.py code will take responsible for modifying the config.py, running the worker for the tests and invoking the pytests, and it will erase all the changes it made when exits. In addition, I also added the line to run the worker in the ./travis.yml and the docstrings to the make_celery function and the ContextTask class.

adarshp:
For now, we are going with the builtin concurrent.futures Python module, via the Flask-executor plugin. We may revisit this later as we attempt to scale up.

PR #137, opened 2018-12-14
Model analysis

pauldhein:
## Description
Implementation of the first variance-based sensitivity analysis method. This PR includes the beginning of a variance method class hierarchy for performing different forms of sensitivity analysis. Further forms of sensitivity analysis such as variogram and emulation methods will use the same style of class hierarchy. Also included in this PR is the Python data files needed to run sensitivity analysis

## Todo
- Implement visualization tools for output variance given inputs
- Implement sensitivity analysis pipeline for use with `ProgramAnalysisGraph`s


adarshp:
@skdebray @pratikbhd There are a couple of refactorings of PA code that I'd like your review(s) on:

1. I switched to using f-strings in some parts of `genCode.py`, and removed unnecessary calls to `str.format`.
2. I changed some code in `genCode.py` and `genPGM.py` so that the lambda functions are more compactly represented in the generated `lambdas.py` file, so that instead of

```python
def CROP_YIELD__lambda__CONSISTENCY_0():
    CONSISTENCY = 64.0
    return CONSISTENCY
```

the generated `lambdas.py` file will now have

```python
def CROP_YIELD__lambda__CONSISTENCY_0():
    return 64.0
```

3. The line 
```python
import math
```
is now at the top of each `lambdas.py` file.

4. I modified `scopes.py` to handle dicts produced by `genPGM.py` that may or may not have a non-empty-value for the `start` key. This fixes the issue I mentioned in https://github.com/ml4ai/delphi/pull/113#discussion_r234463474 (well, partially - when there are multiple subroutines, the first one will be taken as the root in `scopes.py` - see below) https://github.com/ml4ai/delphi/blob/02075a7d90cff3d886b4211fe8bd3dbb4d02d19d/delphi/program_analysis/scopes.py#L62-L70 - probably good for @pauldhein to also review this one.

adarshp:
For the sake of expediency, I'm going to go ahead and merge this into master - I assume that there are no changes that are super controversial here.  I'll leave the commits unsquashed for easy reverts if necessary.

Issue #138, opened 2018-12-15
Malformed lambdas.py


adarshp:
The current version of `lambdas.py` produced for the `PETPT.for` program is malformed. Some of the lambda functions have `math.exp` as a variable name in the list of function arguments, like the one below:

```python
def PETPT__lambda__ALBEDO_1(MSALB, math.exp, XHLAI):
    ALBEDO = (0.23-((0.23-MSALB)*math.exp(-((0.75*XHLAI)))))
    return ALBEDO
```

Upon trying to import the `lambdas.py` module, we get the following error:

```python
>>> import PETPT_lambdas
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/adarsh/Downloads/PETPT_lambdas.py", line 12
    def PETPT__lambda__ALBEDO_1(MSALB, math.exp, XHLAI):
                                           ^
SyntaxError: invalid syntax
```

Looks like you can't have dots (`.`) in the name of arguments in function declarations in Python...

Unless it would break something else, I suggest simply removing the `math.exp` from the argument list in the function declaration.

pratikbhd:
Removing `math.exp` from the argument list will not break anything else. But I feel in doing that, it would also mean removing other built-in functions such as `max` (as in the case of PETPT.for).
I feel it would make more sense to remove just the `math.` section and retaining the `exp` so that we can include other functions as well.

adarshp:
I don't think we need to specify built-in functions such as `max` as inputs to the lambda functions either (it wouldn't be desirable, i.e. preserve the semantics of the program if we left the definition of `max` unconstrained - or in other words, it wouldn't make sense to call the `lambda` function with any other function taking the place of the `max` function)

Re: other functions - let's table this until we are actually dealing with non-built-in functions.

adarshp:
Fixed by #139

PR #139, opened 2018-12-17
Removing functions as arguments to lambda functions

adarshp:
This PR removes Fortran functions (builtin or otherwise) as arguments to lambda functions (see #138 )


adarshp:
Yep, just this one line :) - I'll go ahead and merge it into master.

PR #142, opened 2018-12-19
Suggested renaming of for2py tool

pauldhein:
The purpose of this second PR is to discuss the renaming of our `for2py` translation tool as well as the creation of a `tranlators` directory that could hold future translators we may wish to implement (i.e. `for2julia`).

The code committed below reflects the changed naming schema proposed by @cl4yton, @adarshp, and @pauldhein. @skdebray please share your thoughts on whether this naming scheme is appropriate or what you would prefer.


skdebray:
The renaming seems very reasonable to me.

cl4yton:
Looks like we have unanimous agreement, please proceed!

PR #151, opened 2019-01-08
File handling

pratikbhd:
Progress on handling file I/O. Handling list-directed read is still work in progress and modifications to the pgm and lambdas are required.


pratikbhd:
@adarshp Okay, I'll do that. Also for future reference, is `tests/test_program_analysis.py` the script to where we add various tests for program analysis so that it calculates the code coverage and performs Travis CI tests?

adarshp:
Yes, that's right. 

If `test_program_analysis.py` gets too large in the future, we might split it into multiple modules.

pratikbhd:
@adarshp Okay got it. Thanks!

PR #158, opened 2019-01-27
bugfix related to weird IF decision block

pratikbhd:
This is a quick fix to the issue with the IF blocks in the decision block.
It seems that the __decision__ tag was still under development when the code for this was written. I think the information for this is not present in the GrFN spec as well. Might have to talk about this to come to an understanding of how/when this is handled.


adarshp:
from PET.for: 

```fortran
! EO      Potential evapotranspiration rate (mm/d)
```

The underlying identifiers are unique (and show up in the GrFN as `EO_1`, `EO_2`, etc.). I just dropped the indices to reduce visual clutter.

cl4yton:
Got it, thanks!

PR #159, opened 2019-01-27
Grfn (version m3) updates

cl4yton:
This pull request makes the following significant updates to the GrFN specification.

@skdebray @pratikbhd @hlim1 @pauldhein: please read through everything carefully to make sure I have (1) captured the discussions so far, (2) am being consistent, and (3) that aspects of the spec that affect you are clear and work. 

Please note anything we should capture as a FUTURE or TODO -- the distinction here is that "FUTURE" means plans for next (or later) spec iterations, while "TODO" means something that should be reviewed now before release of this version, and either resolved or turned into a FUTURE tag.

The following lists main changes:
- Added explicit version number along with changes list
- Revised introduction, specification conventions and wording of background description of "source code to dynamic system representation"
- Added identifiers: Grounding, Base name, Scope and Namespace paths, Path string representation, Identifier string representation, Identifier gensym, <identifier_spec>
- Updated Variable and Function identifier and reference section to incorporate identifiers and make the naming conventions more consistent
- Update Top-level GrFN specification to include "identifiers" as list of <identifier_spec>; <variable_spec>s will remain specified within functions (now that identifiers have been defined, the separate out some of what had been loaded into variables, so separating identifiers from variable and function names cleans things up a bit...); includes updates to <variable_spec> examples
- Generally cleaned up Function specification discussion


cl4yton:
Hi Everyone, you can directly observe the diff using this url:
https://github.com/ml4ai/delphi/pull/159/files

docs/grfn_spec.md
```diff
@@ -11,96 +20,189 @@ associated extracted comments, links to natural language text (the
...
+
+    CHOICE: Do we declare each identifier separately, or combine them at program analysis time to the a single identifier with aliases.
+    
+    FOR NOW: We will only keep track of a single identifier (the first one encountered by program analysis) but associate any additional `aliases` as the names of any additional identifier introduced in code.
+	
+- "source\_references": To facilitate later grounding inference, we will store a reference to the location within the source code where an identifier is declared, using a `<source_code_reference>`:
+
+	```
+    <source_code_reference> := <string>
```
pauldhein: I thought we were using `::=` to denote a new rule in BNF (as opposed to `:=`)? Is this not a rule statement?

docs/grfn_spec.md
```diff
@@ -11,96 +20,189 @@ associated extracted comments, links to natural language text (the
...
+
+- "aliases": It is possible for multiple identifiers to be used to denote the same program element. How this is done differs across languages, according to scoping rules and assignment. Program analysis modules for each language (e.g., the Fortran `for2py` analyzer) will determine this. One general way to assign more than one identifier to the same program element is through a *simple equality assignment*, e.g.: `y = x` means that a new identifier, `y`, denotes the same program element that `x` does. A simple equality assignment just involve one identifier being equated with another, no other operations are applied; if other operations are applied (e.g., `y = x + 1`), then this is a *new* identifier as it does not represent the original value of `x` but a modification of it.
```
skdebray: "aliases": ... "Program analysis modules for each language (e.g., the Fortran `for2py` analyzer) will determine this."  Note that once we have pointers (and DSSAT seems to have them) it can become impossible in general to determine all aliases statically.  Suggest making this a FUTURE issue.

docs/grfn_spec.md
```diff
@@ -11,96 +20,189 @@ associated extracted comments, links to natural language text (the
...
+
+### Scope and namespace paths
+
+Identifiers may have the same `base_name` (as it appears in source code) but be distinguished by either (or both) the [scope](https://en.wikipedia.org/wiki/Scope_(computer_science)) and [namespace](https://en.wikipedia.org/wiki/Namespace) within which they are defined in the source code.
+
+Each source language has its own rules for specifying `scope` and `namespace`, and it will be the responsibility of each program analysis module (e.g., Fortran `for2py`) to identify the hierarchical structure of the context that uniquely identifies the specific `scope` and/or `namespace` within which an identifier `base_name` is defined. However, generally `scope`s and `namespace`s may be defined hierarchically, such that the name for each level of the hierarchy taken together uniquely define the context. A `path` of names appears to be sufficient to generally represent the hierarchical context for either a specific `scope` or `namespace`. In general, a names for a `path` are listed in order from general (highest level in the hierarchy) to specific.
```
pauldhein: > In general, ~~a~~ names for a `path` are

docs/grfn_spec.md
```diff
@@ -11,96 +20,189 @@ associated extracted comments, links to natural language text (the
...
+            baz.py
+    ```
+    would be the uniquely specified by the following path:
+    ```
+    ["foo", "bar", "baz"]
+    ```
+
+	In the case of declared namespaces, the namespace declaration will determine the path (which may only consist of one string name).
+
+    Again, it is not necessary within GrFN to independently declare a `namesapce`; like the `<scope_path>`, we specify the `<namespace_path>` within an identifier declaration as a list strings under the "namespace" attribute in the identifie declaration:
```
pauldhein: `namesapce` --> `namespace`

attribute in the identifie declaration --> attribute in the identifier declaration

docs/grfn_spec.md
```diff
@@ -114,93 +216,100 @@ And the second:
...
-function is defined. The `<enclosing_context>` represents the source
-code environment context within which the function is defined. When a
-function is declared at the \"top level\" (as will often be the case for
-container functions), the `<enclosing_context>` is just the \"top
-level\" and so is empty. Assigns, conditions and loop\_plates, however,
-will often be declared within another source code function. In those
-cases, the `<enclosing_context>` capture the enclosing function name.
-For example, for an assignment within the UPDATE\_EST function, the
-`<enclosing_context>` is \"UPDATE\_EST\".
+    ```python
```
skdebray: Misformatted

docs/grfn_spec.md
```diff
@@ -11,96 +20,189 @@ associated extracted comments, links to natural language text (the
...
+
+### Identifier string
+
+Identifiers are uniquely defined by their `base_name`, `scope`, and `namespace`. It will be convenient to refer to any identifier using a single string, outside of the identifier specification declaration. We define an `<identifier_string>` by combining the `namespace`, `scope` and `base_name` (in that order) within a single string by separating the `<namespace_path_string>`, `<scope_path_string>` and `<base_name>` by double-colons:
+
+    <identifier_string> ::= "<namespace_path_string>::<scope_path_string>::<base_name>"
+
+### Identifier gensym
+
+One of the outputs of program analysis is a functionally equivalent version of the original source code (e.g., Fortran) and lambda functions, both expressed in the Python (as the intermediate target language). All identifiers in the output Python must match identifiers in GrFN. Since capturing the semantics (particularly the namespace and scope context) results in a representation that does not appear to be consistently expressible in legal Python output, we will use `gensym`s that may be represented (generally more compactly) as legal Python names and associated uniquely with identifiers.
```
pauldhein: > both expressed in ~~the~~ Python

docs/grfn_spec.md
```diff
@@ -210,35 +319,41 @@ JSON attribute-value list, with the following schema definition:
...
     <grfn_spec>[attrval] ::=
         "start": <string>
-        "name" : <string>
-        "dateCreated" : <string>
+        "source" : <string>
+        "date_created" : <string>
+        "identifiers" : list of <identifier_spec>
         "functions" : list of <function_spec>
 
+There will be one GrFN spec file corresponding to each source code file
```
skdebray: I wonder whether it's necessary to require a separate GrFN file for each individual source file.  Instead, would it be OK to have a single GrFN spec for a program?  E.g.: from earlier discussions about the SimpleModular code I had the impression that @pauldhein would prefer to have a single GrFN file instead of four separate files corresponding to the four different source files.

docs/grfn_spec.md
```diff
@@ -11,96 +20,189 @@ associated extracted comments, links to natural language text (the
...
+
+Each identifier will be associated one-to-one with a unique `<gensym>`.
 
-A top level source variable named ABSORPTION would then simply have the
-`<source_variable_name>` as the string making up the
-\<variable\_name\>:
+### Identifier specification
 
-    <variable_name> ::= "ABSORPTION"
+Each identifier within a GrFN specification will have a single `<identifier_spec>` declaration, and `<identifier_string>`s are then used to denote uses of the identifier elsewhere in the GrFN spec (e.g., in variable and function specs). An identifier will be declared in the GrFN spec JSON by the following attribute-value list:
```
pauldhein: > and `<identifier_string>`s are then used to denote uses of the identifier elsewhere in the GrFN spec (e.g., in variable and function specs).

This portion seems out of place. This section is supposed to be about the identifier specification. I think we should move this important piece of information to the end of the Identifier string section, so that we capture the definition and use case of the identifier string object in one section.

docs/grfn_spec.md
pauldhein: I agree with @skdebray on this point. I think it is more natural to have a GrFN spec correspond to a program. This also helps shed light on the issue that some programs have no single "start" point and indeed may have no start point at all (unless you consider any subroutine as a start point).

docs/grfn_spec.md
```diff
@@ -210,35 +319,41 @@ JSON attribute-value list, with the following schema definition:
...
 - It may also be desirable to add an attribute to represent
   the program analysis code version used to generate the GrFN (as
   presumably the program analysis code could evolve and have different
   properties) \-- although \"dateCreated\" may be sufficient.
 
-A (partial) example instance of a JSON attribute-value list generated
-following the `<grfn_spec>`:
+NOTE: variables are not declared at the top-level `<grfn_spec>`, but will be defined in `<function_spec>`s, described below.
+
+A (partial) example instance of the JSON generated for a `<grfn_spec>` of an analyzed file in the path \'crop\_system/yield/crop\_yield.py\' is:
```
pauldhein: Is it the case that there is going to be one entry in the "functions" section for every entry in "identifiers"? If so, then I suggest we use a hash-map style datastructure pointed to by functions with the "identifiers" as the keys and the data that would have been stored in "functions" as the objects.

docs/grfn_spec.md
```diff
@@ -435,34 +548,26 @@ parsed to extract the actual value according to its data type).:
...
+FOR NOW: have the lambda function reference the source code that does the computation, in the translated Python generated by program analysis. Any variables that are involved in the computation must be listed in the \"source\" list of variables (\<variable\_name\> references) in the \<function\_assign\_spec\>.
 
-To start, the `<source_code_reference>` string could just be the line
-number or a tuple denoting the range of line numbers over which the
-Python source code for the corresponding operations are defined.
+As noted above, due to the more semantically rich identifier specification and `<identifier_string>` representation, it is not straightforward to use the `<identifier_string>` as the python symbol in the translated Python generated by program analysis. Instead, function and variable identifiers will be represented int he generated Python using their gensym. For debugging and visualization purposes, the generated Python code may be displayed with `<identifier_string>` (or some version that is closer to legal Python naming, although in general it does not appear to be possible to create \"safe\" Python names directly from `<identifier_string>`s).
```
pauldhein: > represented ~~int he~~ **in the** generated Python

docs/grfn_spec.md
```diff
@@ -11,96 +20,189 @@ associated extracted comments, links to natural language text (the
...
+An identifier is a symbol used to uniquely identify a program element in code, where a *program element* is a 
+- variable (or constant)
+- function
+- type (class)
+
+More than one identifier can be used to denote the same program element, but an identifier can only be associated with one program element at a time.
+
+### Grounding
+
+Identifiers play a key role in connecting the model as implemented in source code to the target domain that it models. *Grounding* is the task of inferring what aspect of target domain a program element may correspond to, and identifiers, by their (base) name(s), their declaration and use (where they occur in code -- `scope` and `namespace`), and the doc and comment strings that occur around them, provide clues to what they are intended to refer to. For this reason, we need to associate with identifiers several pieces of information. This information will be collected during program analysis and associated with the identifier declaration:
```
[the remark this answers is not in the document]
cl4yton: Agree! I have reworded (not yet pushed)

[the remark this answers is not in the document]
cl4yton: *Grounding* is the task of inferring what aspect of target domain a program element may correspond to. Identifiers, by their (base) name(s), their declaration and use (i.e., where they occur in code, through their `scope` and `namespace`), and the doc and comment strings that occur around them, provide clues to what program elements are intended to represent in the target domain.

docs/grfn_spec.md
```diff
@@ -11,96 +20,189 @@ associated extracted comments, links to natural language text (the
...
+
+- "aliases": It is possible for multiple identifiers to be used to denote the same program element. How this is done differs across languages, according to scoping rules and assignment. Program analysis modules for each language (e.g., the Fortran `for2py` analyzer) will determine this. One general way to assign more than one identifier to the same program element is through a *simple equality assignment*, e.g.: `y = x` means that a new identifier, `y`, denotes the same program element that `x` does. A simple equality assignment just involve one identifier being equated with another, no other operations are applied; if other operations are applied (e.g., `y = x + 1`), then this is a *new* identifier as it does not represent the original value of `x` but a modification of it.
+
+    CHOICE: Do we declare each identifier separately, or combine them at program analysis time to the a single identifier with aliases.
```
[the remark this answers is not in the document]
cl4yton: Fixed:
> Do we declare each identifier separately, or combine them at program analysis time to treat them as a single identifier with aliases?

docs/grfn_spec.md
cl4yton: Good point. Added this note:
> FUTURE: Note that once we consider pointers (e.g., DSSAT has some), it can become impossible in general to determine all aliases *statically*.

docs/grfn_spec.md
cl4yton: Good points. Here is what I've changed the text to (for both the "source" and "start" attributes):

>There may be a single GrFN spec file for multiple source code files.
>    FOR NOW: The main target of a GrFN spec file is all of the source code files involved in defining a program.
>    For this reason, the \"source\" attribute is a list of one or more source code files. The string identifying a source file is represented the same way as a `<namespace_path>` (as described above), except that the final name (the name of the file itself) _will_ include the file extension.
>
> It is also the case that there may be multiple \"start\" points (or none at all) for a given program. For this reason, the \"start\" attribute is a list of zero or more names of the entry point(s) of the (Fortran) source code (for example, the PROGRAM module). In the absence of any entry point, this value will be an empty list: `[]`.

cl4yton: Here is the updated `<grfn_spec>` top level:

```
    <grfn_spec>[attrval] ::=
        "source" : list of <string>
        "start": list of <string>
        "date_created" : <string>
        "identifiers" : list of <identifier_spec>
        "functions" : list of <function_spec>
```

docs/grfn_spec.md
cl4yton: Agreed, I have cut this, and slightly reworded the description in the above Identifier string section to:

>It will be convenient to refer unambiguously to any identifier using a single string, outside of the identifier specification declaration (defined below).

and

>`<identifier_string>`s will be used to denote identifiers as they are used in variable and function specifications (described below).

docs/grfn_spec.md
cl4yton: Identifiers are just names, they could be used to denote variables or functions. The identifier that is associated with a variable needs to be declared as such. Such declarations, as needed for use in functions, are therefore associated directly with the functions. Let discuss this if it's not clear (possibly I'm not understanding your concern).

docs/grfn_spec.md
```diff
@@ -114,93 +216,100 @@ And the second:
...
+    
+    	"assign$CROP_YIELD::UPDATE_EST::YIELD_EST"
+    
+	If this assignment takes place in the function UPDATE\_EST and the 
+	namespace CROP\_YIELD, then the full `<identifier_string>` of the 
+	function identifier would be:
 
--   **Condition**: A condition within the function UPDATE\_EST assigning
-    the (inferred) boolean variaaible IF\_1:
+        "CROP_YIELD::UPDATE_EST::assign$CROP_YIELD::UPDATE_EST::YIELD_EST"
```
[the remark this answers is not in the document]
cl4yton: Will make example where the identifier begin assigned is in a different namespace/scope pair from where function is defined.

docs/grfn_spec.md
```diff
@@ -11,96 +20,189 @@ associated extracted comments, links to natural language text (the
...
+
+    Because an identifier may have associated aliases, the "source\_references" will be a list of one or more `<source_code_reference>`, with one `<source_code_reference>` representing the location in source code of the initial identifier declaration, and then an additional `<source_code_reference>` for each time a new alias is initially declared (e.g., through an assignment). The order of these `<source_code_reference>`s will correspond the order of the introduction/declaration of the aliases in the source code (from the perspective of program analysis). (The reason for this is that the initial introduction of the alias is more likely to have associated relevant comments that might provide information about the identifier's grounding.)
+
+### Base name
+
+The `base_name` is intended to correspond to the identifier token name as it appears in the source language (e.g., Fortran). The `<lang_string>` is itself a string
+
+    <lang_string> ::= <string>
```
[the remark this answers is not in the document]
cl4yton: After thinking about this more, I completely agree with you, I'm dropping `<lang_string>` and replacing with `<base_name>`.

PR #160, opened 2019-01-28
Arrays

hlim1:
Codes for creating single-dimension, multidimensional, and array accessors (get_ and set_) translation are added and produce to python file as expected based on the test files provided. It is ready to merge into the master.


skdebray:
@hlim1 Did you run all of the test programs `arrays-basic-o[1-8].f` to confirm that the generated Python code behaves the same as the original Fortran?

I ask because one of those tests (`arrays-basic-08.f`) contains a GOTO, which I don't believe we currently handle (I've been meaning to fix that program but haven't gotten around to it).  This makes me think that this code was not tested on all of those tests?  Can you please confirm?

hlim1:
@skdebray <https://github.com/skdebray> I have not tested #8 at least in the original form. There is GOTO that we have not handled, so I removed the GOTO and modified the test case to have only the DO and DO WHILE loops. I’ve ran the other test cases 1 - 7. Moreover, since the file I/O was not handled in the ‘array’ branch, I’ve modified the test cases to have print statement rather than write. However, there is a minor fix that needs to be done, which I have not caught purely because of my inattention to the detail. I would like to re-create the ‘array’ branch and make a fix under it. The code that was merged should not affect any other code in the master branch.

skdebray:
@hlim1 Thanks for the clarification -- that makes sense.  Please go ahead and restore the ranch and make the fix you mentioned.

Also (and purely for curiosity's sake): what happens when you run the code on the test program with the GOTO?

skdebray:
@hlim1 Thanks for checking.  Go ahead and commit the code and send a PR once you've fixed the issue you were thinking of.  At our meeting tomorrow, let's discuss how we should deal with unhandled source constructs.

PR #161, opened 2019-01-29
Arrays

hlim1:
Fixed bug: Code was not handling the case of a form where one array using another array's value as an index, for example, array1(array2(i)). All produced python programs behave the same as the tests arrays-basics-01 to 07


tests/data/arrays/arrays-basic-05.f
```diff
@@ -16,7 +16,7 @@ program main
 
       ! write out the array
       do i = -3, 1
-          write (*,10) A(i,-4), A(i,-3), A(i,-2), A(i,-1), A(i,0)
+          write (*,10) A(-1,-4), A(i,-3), A(i,-2), A(i,-1), A(i,0)
```
skdebray: @hlim1 I don't understand this edit.

hlim1:
@skdebray I was testing on the case when an array is a multidimensional and the left side of the comma  (the first dimension) has an operator with literal value, so I made a slight change A (i, -4) to A (-1, -4) but then I forgot to reserve back to the original form. My mistake, again. It doesn’t have any other meaning other than for case testing.

skdebray:
OK thanks.  Please restore the original version and push your update.  --s

Issue #166, opened 2019-02-01
Type misnaming in generated GrFN JSON


pauldhein:
Functions in the GrFN JSON that are `condition`s or `decision`s should have their types labeled as such. Currently our `pgm.json` for `crop_yield.f` has the following function definitions:
```json
{
      "name": "UPDATE_EST__condition__IF_1_0",
      "type": "assign",
},
...
{
      "name": "UPDATE_EST__decision__YIELD_EST_0",
      "type": "assign",
},
```

We need to change these definitions to the following:
```json
{
      "name": "UPDATE_EST__condition__IF_1_0",
      "type": "condition",
},
...
{
      "name": "UPDATE_EST__decision__YIELD_EST_0",
      "type": "decision",
},
```
The `for2py` module should be updated to reflect this change so that future `condition`s and `decision`s are identified with their types properly as opposed to being identifier with `assign`.

hlim1:
@pauldhein I fixed the issue in the genPGM.py and ran the test. The output shows the correct type names with "condition" and "decision". I just requested the issue_#166 branch that has a fix to be merged with the master.

skdebray:
@pauldhein let me know if anything further needs to be done for this.  If not I'll go ahead and close this issue.

pauldhein:
This issue is ready to be closed

Issue #167, opened 2019-02-02
Travis test failing


adarshp:
@pratikbhd One of the changes introduced in #165 seems to make the automated test suit fail (https://travis-ci.org/ml4ai/delphi/builds/487867022) - could you please submit a PR that fixes it?

skdebray:
@adarshp If you're satisfied with the fix I'll go ahead and close this issue.

adarshp:
Yes, I am! I'll close it now.

Closed by #168 .

PR #171, opened 2019-02-07
Issue #166

hlim1:
Handled type miss naming issue (#166 ). Tested with the crop_yield.f and produced the type name correctly as expected.


skdebray:
@hlim1 It looks like a travis build is failing for the same reason as issue #167, which has to do with the OpenFortranParser jar files (check with @pratikbhd).  Can you please fix this?

PR #183, opened 2019-02-21
Program analysis graph wiring update

adarshp:
This PR implements a wiring update so that a CAG is produced with the right link structure from the simple Fortran program below:

```fortran
      program main
      implicit none
      integer i, x, y
      x = 1
      y = 0

      if (x .eq. 1) then
          y = 1
      end if

      end program main
```

The resulting computation graph and CAG are shown below:

screen shot 2019-02-21 at 12 10 32 pm

screen shot 2019-02-21 at 12 11 06 pm

FYI @skdebray

### DIFF
--- a/delphi/GrFN/ProgramAnalysisGraph.py
+++ b/delphi/GrFN/ProgramAnalysisGraph.py
@@ -32,15 +32,15 @@ class ProgramAnalysisGraph(nx.DiGraph):
 
             # Otherwise append the predecessor function list
             elif n.attr["label"] == "__decision__":
-                preds = A.predecessors(n)
                 if_var, = [
                     n
-                    for n in preds
-                    if list(A.predecessors(n)) != []
-                    and list(A.predecessors(n))[0].attr["label"]
-                    == "__condition__"
+                    for n in A.predecessors(n)
+                    if "IF" in n
                 ]
                 condition_fn, = A.predecessors(if_var)
+                for pred in A.predecessors(condition_fn):
+                    self.add_edge(pred.attr["cag_label"], oname)
+
                 condition_fn = condition_fn[: condition_fn.rfind("__")]
                 condition_lambda = condition_fn.replace("condition", "lambda")
                 onode["condition_fn"] = getattr(lambdas, condition_lambda)
@@ -52,6 +52,8 @@ class ProgramAnalysisGraph(nx.DiGraph):
                 for i in A.predecessors(n):
                     iname = i.attr["cag_label"]
                     self.add_edge(iname, oname)
+
+
 
     def add_variable_node(self, n):
         """ Add a variable node to the CAG. """
@@ -162,7 +164,6 @@ class ProgramAnalysisGraph(nx.DiGraph):
         """ Initialize the value of input function nodes in the CAG."""
 
         for n in self.nodes():
-            # if self.nodes[n].get("init_fn") is not None:
             if n in self.input_functions:
                 self.nodes[n]["value"] = self.nodes[n]["init_fn"]()
         self.update()
--- a/delphi/GrFN/scopes.py
+++ b/delphi/GrFN/scopes.py
@@ -408,7 +408,7 @@ class Node(metaclass=ABCMeta):
         return f"{self.name}_{self.index}__{self.scope}"
 
     def get_label(self):
-        return self.name
+        return '_'.join((self.name, str(self.index)))
 
 
 class FuncVariableNode(Node):


skdebray:
@adarshp Thank you.  However, in light of the discussion today where it was decided that CAGs should capture only explicit data dependencies, it seems to me that there should not be an edge from x to y?

Sorry for making things difficult. :-(

adarshp:
Oh! I totally misunderstood the discussion today then :P - I suppose my intuitive criterion of whether there should be a link in a CAG between two nodes X and Y is along the lines of _Can the value of X affect the value of Y (assuming X and Y are the only two nodes in the CAG and are variables that represent some real-world physical quantity)?_ If yes, then there should be a link from X to Y.

I'm certainly not saying that we will for sure be able to capture all the implicit data dependencies when constructing a CAG from the GrFN representation, but I think it's worthwhile to get some of the low-hanging fruit like this one...

@cl4yton Thoughts?

skdebray:
@adarshp @cl4yton There are definitely algorithms for capturing dependencies that arise out of the control-flow structure of programs.  In other words they allow you to make assertions of the form "There is an edge between X and Y in the CAG iff the program's control flow graph (or AST) satisfies some well-defined structural properties".  This adds more work to the CAG construction process, but in return allows you to make that sort of assertion.  So IMO the real question is: what sort of soundness assertion do we need to make about the exact nature of the dependencies captured by a CAG.

adarshp:
> @adarshp @cl4yton There are definitely algorithms for capturing dependencies that arise out of the control-flow structure of programs. In other words they allow you to make assertions of the form "There is an edge between X and Y in the CAG iff the program's control flow graph (or AST) satisfies some well-defined structural properties". This adds more work to the CAG construction process, but in return allows you to make that sort of assertion. So IMO the real question is: what sort of soundness assertion do we need to make about the exact nature of the dependencies captured by a CAG.

I think that question warrants a longer discussion - let's take it up at the next PA meeting? In the meantime, would you be okay with me merging this into master, @skdebray ?

cl4yton:
These are good examples and also highlight that I still need to digest the distinction between data flow and control flow and the importance of maintaining a distinction (whether it is or is not important).

@skdebray : why do you propose that in the code example y is not a direct function of x? Is it because _locally_ the two assignments to y do not directly access x? I do see that point. However, as soon as we move further beyond the conditional (i.e., as soon as we possibly next "use" or access y), any access to the value of y is getting a value that is what it is as a direct result of the state of x. So simply leaving leaving x and y unconnected seems to loose information.

You note that possibly we want to be careful here b/c inferring relationships through control flow could get hard. I'd like to see some more cases where this might happen. In this case, there's no hard inference to be made: due to the work already done by PA to construct the graph representation of the conditional, we get the direct connection of x to y. As I see it, the construction of the condition in the GrFN function network is what converts the control structure to _become_ represented as data-flow. @adarshp and @pauldhein simply "read off" the direct connection. I hypothesize that this will be the case for any control flow we translate to GrFN -- the goal of translating to GrFN should be to make such direct influences as a function of control flow aparent.

This does mean I'm now going back on what I naively thought was the solution in the meeting, to simply say CAG == data flow.

The question I have is then: does doing things this way introduce a disconnect with how you on the program analysis side would like to naturally view things? I take this seriously as overall we seek a tool that serves multiple kinds of analysts. Which brings me back to the possibility that we will want several well-defined kinds of "views", each of which should include a notion of one or more analysis uses that they serve.

For model analysis, from the perspective of causal influence (i.e., how one variable is affected directly by others), I do think directly representing influence through control structures (on top of direct data flow) is the right thing to consider; here the key challenge is being clear by what we mean by _direct_ influence. In the condition example above, I think "directness" is determined by considering what could have led to y having the value it does the next time we go to use y. Also, this doesn't just mean that _any_ prior variable in the chain of influences directly affects y: if there was a z that affects the state of x, then z does have influence on y, but only through x. So z does not directly affect y, but x is consider a proximal influencer of y. This is exactly the concept of a Markov blanket (https://en.wikipedia.org/wiki/Markov_blanket). If we define a CAG this way, the CAG still abstracts away what exactly the function is between x and y, e.g., whether (1) y is assigned one value or another because of the state of x, or (2) y is an assignment as a direct function of the value of x. But in either case, x is still directly implicated in setting the state of y before y gets "used" by anything else.

Thoughts?

skdebray:
@adarshp : Re "In the meantime, would you be okay with me merging this into master" -- absolutely.  I didn't mean to hold that up, my apologies.

adarshp:
@skdebray Thanks! Will merge now.

skdebray:
@cl4yton   Re: "*why do you propose that in the code example y is not a direct function of x? Is it because _locally_ the two assignments to y do not directly access x?*"

I didn't mean to suggest that `y` is not *dependent* on `x`; simply that it is not *data dependent* on `x`.  But maybe this merits a digression to how we define the notion of "data dependence".

The PL literature introduces two notions of dependencies: *data dependencies* and *control dependencies*.  A variable `x` is *directly data dependent* on a variable `y` if there is a program statement of the form
```
x = ... y ...
```
The general notion of "data dependence" is then the transitive closure of this notion of direct data dependence.  

A statement `S` is *control dependent* on a statement `T` if the result of executing `T` determines whether or not `S` is executed.  

In the tiny toy example that started this discussion thread, namely,
```
      program main
      implicit none
      integer i, x, y
      x = 1
      y = 0

      if (x .eq. 1) then
          y = 1
      end if

      end program main
```
there is no direct data dependence between `x` and `y`, therefore no data dependence between `x` and `y`.  There is definitely information flow from `x` to `y`, in that the value of `y` depends on the value of `x`, but this happens through a control dependence (because whether or not the assignment `y = 1` is executed depends on the value of `x`), not a data dependence.

Much to discuss, but I have to run to a meeting... let me know if this helps clarify where I'm coming from.

skdebray:
@cl4yton @adarshp I have a question about reading CAGs (because I may not be reading them correctly).  Consider the following program:
```
      subroutine foo(y, retval)
      implicit none
      integer x, y, z, w, retval

      x = 0
      z = 0
      if (y .eq. 1) then
          x = 1
      else
          z = 1
      end if
      
      z = 0

      if (z .eq. 0) then
          w = 1
      else
          w = 23
      end if

      retval = w
      end subroutine foo
```
The CAG I get for this is 
image

This seems to have a chain of dependencies that suggests that the input argument `y` can influence the output value `retval`; in practice that's not the case -- `retval` does not depend on `y`.  But the computational graph for this program correctly shows that `retval` does not depend on `y`.  How should I be reading CAGs?

adarshp:
Right, this is an interesting example, and I'm not 100% sure what the right way to construct the CAG is here. 

You are right in that in practice, `retval` does not depend on `y`, due to the presence of the line `z=0`. Two thoughts I have:

1. If `x`, `y`, `z`, and `retval` are all grounded to 'real-world' variables, then the CAG structure seems reasonable to me - basically, `retval`depends on `z`, and for this particular instance, `z` is fixed at some point within the code, but overall if we consider the CAG to represent some kind of physical model, then I think the dependency structure should still be there.
2. As you point out, the computational graph (shown below) is indeed disconnected.

screen shot 2019-02-21 at 6 08 00 pm

This reminds me of how 'interventions' are represented in probabilistic graphical models - here is an example from Koller & Friedman's _Probabilistic Graphical Methods_:

screen shot 2019-02-21 at 6 31 43 pm

In this example, the Bayes network on the left is 'mutilated' by fixing the value of 'Grade' - this effectively removes all incoming links into the 'Grade' node.

So, basically, my mental model is something like this now: while the CAG is showing something analogous to the original Bayes network in (b), the computational graph view actually exposes that the variable `z` is actually being intervened on in this case.

Perhaps there should be another 'view' of the CAG that shows the intervention structure, that will also be more faithful to what the dependency structure looks like in the computation graph?

skdebray:
I don't know whether that sort of mutilation is common in other situations.  In code, I can see it happening via some common code idioms, e.g., where a variable `i` is used as the loop index in two different loops:
```f
      subroutine sumprod(x,sum,prod)
      implicit none
      integer x, i, sum, prod

      sum = 0
      do i = 1, 1000        ! sum of the integers up to x
          if (i .le. x) then    ! a poor man's while loop
             sum = sum + i
          end if
      end do

      prod = 1
      do i = 1, 1000        ! product of the integers up to x
          if (i .le. x) then    ! a poor man's while loop
              prod = prod * i
          end if
      end do
      end subroutine sumprod

      program main
      implicit none
      integer sum, prod
      call sumprod(7, sum, prod)
 10   format(I5,X,I5)
      write(*,10) sum, prod

      end program main
```
This seems to me to be a common sort of code pattern.  Since the index is reinitialized at the beginning of each loop, making the loops independent.  I wonder whether the CAG would have spurious dependency edges between things computed in the first loop and those in the second?  (I got an error when I tried to run this program on codeExplorer.  I'll try and take a look tomorrow if possible.)

PR #191, opened 2019-02-26
Derived types

hlim1:
The basics of derived types handling, such as derived type declaration with scalar type field variables and single-dimension array, declare derived type variable within main, stdout via write and print, etc. However, there the test 4, which holds nested derived type needs to be handled, but it is not only limited to the derived type as it is more of translate.py reading XML AST in issue. This will be dealt in the separate branch.


adarshp:
Did we decide not to go with the `dataclass` route for derived types?

skdebray:
@adarshp I had originally planned to go with dataclasses, which would simplify the generated code by eliminating the need to generate standard methods like `__init__()`.  But it turns out that in order to handle derived types that have fields that are aggregates, such as arrays or other derived types, we have to generate the `__init__()` methods explicitly anyway (see [this document](https://docs.google.com/document/d/146RZkX9nY-FiOQV6eC4y6n4NvTBz9UA9iTIed8NJpss/edit?usp=sharing)), at which point it seemed simplest to go with regular classes.

adarshp:
Maybe I am missing something, but it looks to me like dataclasses can handle initialization and arrays - the following script:

```python
from dataclasses import dataclass
from delphi.translators.for2py.scripts.for2py_arrays import Array

@dataclass
class MyClass:
    x: int = 1
    y: Array = Array(["type1", "type2"], [(1,10), (2, 5)])

classInstance = MyClass()
print(classInstance.x)
print(classInstance.y)
print(classInstance.y.bounds())
```

gives me the output I expect:

```python
1
<delphi.translators.for2py.scripts.for2py_arrays.Array object at 0x1032b60f0>
[(1, 10), (2, 5)]
```

skdebray:
@adarshp You're right.  My understanding of dataclasses was incorrect -- I had mistakenly thought that they couldn't handle certain constructs.  Your comment, and my subsequent experiments, show that that's not true.

Given this, I think it makes sense to re-do our handling of derived types to use dataclasses.  I'd like to discuss the priorities for the various tasks in the queue so that we're all on the same page.  Are we having the regular weekly PA meeting tomorrow morning?

PR #196, opened 2019-03-01
Removing necessity for List wrapping

adarshp:
This PR removes the necessity for wrapping Python type hints with List for Python programs that are not automatically generated from Fortran programs (this way we can extract structure from some simple Python functions that don't have return values)

Test program:

```python
from typing import List

def add():
    x: int = 1
    y: int = 2
    z: int = x+y

add()
```

The generated `lambdas.py` file:

```python
import math

def add__lambda__x_0():
    return 1

def add__lambda__y_0():
    return 2

def add__lambda__z_0(y, x):
    return (x+y)
```

Computational graph:

add

Generated GrFN JSON:

```json
{
  "functions": [
    {
      "name": "add__assign__x_0",
      "type": "assign",
      "sources": [],
      "body": {
        "type": "literal",
        "dtype": "integer",
        "value": "1"
      },
      "target": "x"
    },
    {
      "name": "add__assign__y_0",
      "type": "assign",
      "sources": [],
      "body": {
        "type": "literal",
        "dtype": "integer",
        "value": "2"
      },
      "target": "y"
    },
    {
      "name": "add__assign__z_0",
      "type": "assign",
      "sources": [
        {
          "name": "x",
          "type": "variable"
        },
        {
          "name": "y",
          "type": "variable"
        }
      ],
      "body": [
        {
          "type": "lambda",
          "name": "add__lambda__z_0",
          "reference": 6
        }
      ],
      "target": "z"
    },
    {
      "name": "add",
      "type": "container",
      "input": [],
      "variables": [
        {
          "name": "x",
          "domain": "integer"
        },
        {
          "name": "y",
          "domain": "integer"
        },
        {
          "name": "z",
          "domain": "integer"
        }
      ],
      "body": [
        {
          "name": "add__assign__x_0",
          "output": {
            "variable": "x",
            "index": 1
          },
          "input": []
        },
        {
          "name": "add__assign__y_0",
          "output": {
            "variable": "y",
            "index": 1
          },
          "input": []
        },
        {
          "name": "add__assign__z_0",
          "output": {
            "variable": "z",
            "index": 1
          },
          "input": [
            {
              "variable": "x",
              "index": 1
            },
            {
              "variable": "y",
              "index": 1
            }
          ]
        }
      ]
    }
  ],
  "start": "add",
  "name": "grfn.json",
  "dateCreated": "2019-02-28"
}
```

### DIFF
--- a/delphi/translators/for2py/scripts/genCode.py
+++ b/delphi/translators/for2py/scripts/genCode.py
@@ -182,9 +182,7 @@ def genCode(node, state):
 
     # AnnAssign: ('target', 'annotation', 'value', 'simple')
     elif isinstance(node, ast.AnnAssign):
-        codeStr = "{0} = {1}".format(
-            genCode(node.target, state), genCode(node.value, state)
-        )
+        codeStr = genCode(node.value, state)
 
     # Assign: ('targets', 'value')
     elif isinstance(node, ast.Assign):
--- a/delphi/translators/for2py/scripts/genPGM.py
+++ b/delphi/translators/for2py/scripts/genPGM.py
@@ -13,6 +13,15 @@ from typing import List, Dict, Iterable, Optional
 from itertools import chain, product
 import operator
 
+BINOPS = {
+    ast.Add: operator.add,
+    ast.Sub: operator.sub,
+    ast.Mult: operator.mul,
+    ast.Div: operator.truediv,
+    ast.Pow: operator.pow,
+    ast.Eq: operator.eq,
+    ast.LtE: operator.le,
+}
 
 class PGMState:
     def __init__(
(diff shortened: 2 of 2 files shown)


adarshp:
@skdebray this is clearly not a general solution, but at least we can start processing some Python programs with more standard type hints...

skdebray:
@adarshp I just want to make sure I understand what will happen with Fortran code, which in general needs list wrapping.  For example, consider the following Fortran code:

```f
C subroutine add: sets arg3 to the sum of arg1 and arg2
      subroutine add(x, y, z)
      implicit none
      integer x, y, z

      z = x+y
      end subroutine add

      program main
      implicit none
      integer val
      call add(111, 222, val)
 10   format(I5)
      write(*,10) val
      end program main
```

This prints out the value 333, which is the sum of the values 111 and 222 given to the subroutine `add()` as arguments.  Suppose we write analogous Python code but without list wrapping, e.g., something like:

```python
def add(x, y, z):
    z = x+y

def main():
    val = None
    add(111, 222, val)
    print(val)

main()
```

In this case, the Python code prints out `None` -- different from the Fortran code.

How will we handle this sort of situation?

adarshp:
Just to clarify, @skdebray , this PR doesn't change how Fortran will be translated into Python - it just enables an entry point into the middle of the pipeline - i.e. if we were starting with a model originally written in Python instead of Fortran.

skdebray:
@adarshp  Got it, thank you.  :)

PR #197, opened 2019-03-01
Dataclasses

adarshp:
This PR switches from regular classes to dataclasses for Fortran derived types.

@skdebray This seemed like a quick fix, so I couldn't resist :)


skdebray:
:1st_place_medal:  :)

skdebray:
^^^ That looks like a pineapple, but really it's a First Place Medal.

Issue #186, opened 2019-02-21
Better naming schema for lambda functions


pauldhein:
I would like the generated lambdas in the `*_lambdas.py` files to be changed from their current form:
```python
def petpt__lambda__td_0(tmax, tmin):
    return ((0.6*tmax)+(0.4*tmin))

def petpt__lambda__IF_1_0(xhlai):
    return (xhlai<=0.0)

def petpt__lambda__albedo_2(IF_1_0, albedo_0, albedo_1):
    return albedo_0 if IF_1_0 else albedo_1
```
to the following form:
```python
def petpt__assign__td_0(tmax, tmin):
    return ((0.6*tmax)+(0.4*tmin))

def petpt__condition__IF_1_0(xhlai):
    return (xhlai<=0.0)

def petpt__decision__albedo_2(IF_1_0, albedo_0, albedo_1):
    return albedo_0 if IF_1_0 else albedo_1
```
This will allow me to more quickly inspect the lambdas file by searching function names based on the type of function I am searching for. Not only that, but I believe we can condense the JSON specification for `__assign__`, `__condition__`, and `__decision__` function objects by using this format. Currently the outputted JSON for three functions mentioned above is:
```json
{
  "name": "petpt__condition__IF_1_0",
  "type": "condition",
  "target": "IF_1",
  "sources": [
    {
      "name": "xhlai",
      "type": "variable"
    }
  ],
  "body": [
    {
      "type": "lambda",
      "name": "petpt__lambda__IF_1_0",
      "reference": 17
    }
  ]
},
{
  "name": "petpt__assign__albedo_0",
  "type": "assign",
  "sources": [
    {
      "name": "msalb",
      "type": "variable"
    }
  ],
  "body": [
    {
      "type": "lambda",
      "name": "petpt__lambda__albedo_0",
      "reference": 18
    }
  ],
  "target": "albedo"
},
{
  "name": "petpt__decision__albedo_0",
  "type": "decision",
  "target": "albedo",
  "sources": [
    {
      "name": "IF_1_0",
      "type": "variable"
    },
    {
      "name": "albedo_3",
      "type": "variable"
    },
    {
      "name": "albedo_2",
      "type": "variable"
    }
  ]
},
```
and with my proposed format I believe we can condense the output into the following:
```json
{
  "name": "petpt__condition__IF_1_0",
  "type": "condition",
  "target": "IF_1",
  "reference": 17,
  "sources": [
    {
      "name": "xhlai",
[...]

Issue #185, opened 2019-02-21
Missing Lambdas for __decision__ nodes


pauldhein:
Consider the following snippet of code from `PETPT.for`:

```fortran
IF (XHLAI .LE. 0.0) THEN
    ALBEDO = MSALB
ELSE
    ALBEDO = 0.23-(0.23-MSALB)*EXP(-0.75*XHLAI)
ENDIF
```

In our `grfn.json` file we get the following functions from this code:
```json
{
      "name": "petpt__condition__IF_1_0",
      "type": "condition",
      "target": "IF_1",
      "sources": [
        {
          "name": "xhlai",
          "type": "variable"
        }
      ],
      "body": [
        {
          "type": "lambda",
          "name": "petpt__lambda__IF_1_0",
          "reference": 17
        }
      ]
    },
    {
      "name": "petpt__assign__albedo_0",
      "type": "assign",
      "sources": [
        {
          "name": "msalb",
          "type": "variable"
        }
      ],
      "body": [
        {
          "type": "lambda",
          "name": "petpt__lambda__albedo_0",
          "reference": 18
        }
      ],
      "target": "albedo"
    },
    {
      "name": "petpt__assign__albedo_1",
      "type": "assign",
      "sources": [
        {
          "name": "msalb",
          "type": "variable"
        },
        {
          "name": "math.exp",
          "type": "function"
        }
      ],
      "body": [
        {
          "type": "lambda",
          "name": "petpt__lambda__albedo_1",
          "reference": 20
        }
      ],
      "target": "albedo"
    },
    {
      "name": "petpt__decision__msalb_0",
      "type": "decision",
      "target": "msalb",
      "sources": [
        {
          "name": "IF_1_0",
          "type": "variable"
        },
        {
          "name": "msalb_0",
          "type": "variable"
        },
        {
          "name": "msalb_0",
          "type": "variable"
        }
      ]
    },
    {
      "name": "petpt__decision__albedo_0",
      "type": "decision",
      "target": "albedo",
      "sources": [
        {
          "name": "IF_1_0",
          "type": "variable"
        },
        {
          "name": "albedo_3",
          "type": "variable"
        },
        {
          "name": "albedo_2",
          "type": "variable"
        }
      ]
    },
```
A couple necessary corrections:
1. `petpt__decision__msalb_0` should not have been generated. This needs to be removed.
[...]

PR #199, opened 2019-03-04
Changes for Issue 186

pratikbhd:
This PR adds the following changes:

1. The GrFN JSON now does not contain the body key as suggested in #186 
2. The `__lambda__` functions have been renamed to their respective tag names i.e. `__assign__`, `__condition__` and `__decision__`
3. Added a `__lambda__` function for `__decision__` tags as suggested in #185

### DIFF
--- a/delphi/GrFN/ProgramAnalysisGraph.py
+++ b/delphi/GrFN/ProgramAnalysisGraph.py
@@ -42,8 +42,7 @@ class ProgramAnalysisGraph(nx.DiGraph):
                     self.add_edge(pred.attr["cag_label"], oname)
 
                 condition_fn = condition_fn[: condition_fn.rfind("__")]
-                condition_lambda = condition_fn.replace("condition", "lambda")
-                onode["condition_fn"] = getattr(lambdas, condition_lambda)
+                onode["condition_fn"] = getattr(lambdas, condition_fn)
             else:
                 onode["pred_fns"].append(getattr(lambdas, n.attr["lambda_fn"]))
 
--- a/delphi/GrFN/scopes.py
+++ b/delphi/GrFN/scopes.py
@@ -431,8 +431,6 @@ class ActionNode(Node):
         self.shape = "rectangle"
         self.lambda_fn = (
             "_".join((name, idx))
-            .replace("assign", "lambda")
-            .replace("condition", "lambda")
         )
 
     def get_label(self):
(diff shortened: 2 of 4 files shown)


pratikbhd:
@adarshp There seems to be an error coming from `ProgramAnalysisGraph.py` after changing the function names in the `lambda` file from the `__lambda__` form to the `__assign\decision\condition__` form.

adarshp:
`ProgramAnalysisGraph.py` loads lambda functions from the lambdas module by name using `getattr`. Could you please change the names used to load the condition/decision functions as necessary in `ProgramAnalysisGraph.py`? I think that should fix the test.

pratikbhd:
@adarshp Okay, I'll go ahead and do that.

adarshp:
You may need to get rid of these lines as well: https://github.com/ml4ai/delphi/blob/ffa649b92b8f47521d139b250574abca6b60e4de/delphi/GrFN/scopes.py#L434-L435

pratikbhd:
It looks like the naming issue has been resolved but I think there is an index mismatch that is still causing the error.

pauldhein:
@pratikbhd `scopes.py` is definitely not handling variable indexing in an ideal fashion. I am currently changing our handling of variable naming to be more logical, but this update will be coming to `GrFN.py`, not the scopes class. However, I investigated the concern you raised about index handling and I have found that there appears to still be some indexing issues coming from the generated JSON and lambdas. The following shows the problem that is currently causing our tests to fail:

Here is what I found in the freshly generated JSON:
```json
{
  "name": "crop_yield__assign__max_rain_0",
  "type": "assign",
  "reference": 24,
  "sources": [],
  "target": "max_rain",
},
``` 

and here is what I found in the freshly generated Lambdas:
```python
def crop_yield__assign__max_rain_1():
    return 4.0
```

`(scopes | GrFN).py` doesn't seem to be the problem when there is an index discrepancy that still exists between the two generated files.

Furthermore, the generated index problem becomes worse down the line. Checkout the following screenshots for even more troublesome problems:
screen shot 2019-03-04 at 11 15 19 am
screen shot 2019-03-04 at 11 16 07 am

It seems that both of your files are incrementing index counts by 2 instead of by 1 and that they are off pace from one another (one generating odd indexes and the other generating evens). I pushed some code that allows the JSON to be outputted when running `tests/test_program_analysis.py` so that you can verify that results I have shown here. I hope this helps you to debug!

pratikbhd:
@pauldhein Thanks for pointing out the mismatch between the generated json and lambda files. I now realize that the lambda file names were being redundantly generated each time, causing their index to be incremented on each generation. This was certainly because of indexing issues in the GrFN generation code. I think the latest commit should fix that.

Issue #201, opened 2019-03-05
Getting rid of global variables in pyTranslate.py


adarshp:
There are two global variables in `pyTranslate.py` - `imports` and `programName` - that should be better encapsulated - it looks like at least one (`programName`) is causing troubles downstream - the value of this variable is persistent, i.e. when two Fortran programs are translated, the first one with a top-level function call, and the second without, a call to the `programName` function will be included in the Python translation.

For example, this happens in the `test_program_analysis.py` test module - `crop_yield.f` is translated before `PETPT.for`, so the last line of the translation of `PETPT.for` is then set to `crop_yield()`

`crop_yield.f` translation:

```python
import sys
from typing import List
import math
from delphi.translators.scripts.for2py.fortran_format import *
from delphi.translators.scripts.for2py.for2py_arrays import *
from dataclasses import dataclass


def update_est(rain: List[float], total_rain: List[float], yield_est: List[float]):
    total_rain[0] = (total_rain[0] + rain[0])
    if (total_rain[0] <= 40):
        yield_est[0] = (-((((total_rain[0] - 40) ** 2) / 16)) + 100)
    else:
        yield_est[0] = (-(total_rain[0]) + 140)

def crop_yield():
    day: List[int] = [0]
    rain: List[float] = [0.0]
    yield_est: List[float] = [0.0]
    total_rain: List[float] = [0.0]
    max_rain: List[float] = [0.0]
    consistency: List[float] = [0.0]
    absorption: List[float] = [0.0]
    max_rain[0] = 4.0
    consistency[0] = 64.0
    absorption[0] = 0.6
    yield_est[0] = 0
    total_rain[0] = 0
    for day[0] in range(1, 31+1):
        rain[0] = ((-((((day[0] - 16) ** 2) / consistency[0])) + max_rain[0]) * absorption[0])
        update_est(rain, total_rain, yield_est)
        print("Day ", day, " Estimate: ", yield_est)
    print("Crop Yield(%): ", yield_est)

crop_yield()
```

`PETPT.for` translation:

```python
import sys
from typing import List
import math
from delphi.translators.scripts.for2py.fortran_format import *
from delphi.translators.scripts.for2py.for2py_arrays import *
from dataclasses import dataclass


def petpt(msalb: List[float], srad: List[float], tmax: List[float], tmin: List[float], xhlai: List[float], eo: List[float]):
    albedo: List[float] = [0.0]
    eeq: List[float] = [0.0]
[...]

Issue #164, opened 2019-01-31
for2py refactor


pauldhein:
## Changes requested
- Lift all of the files in `delphi/translators/for2py/scripts/` to the `for2py` directory
- Create a directory for the temporary files that are generated when running `autoTranslate`
- Edit the `for2py` code such that all temporary files are created in the temporary file directory
- Add the path to the temporary file directory to the `.gitignore` file

adarshp:
Closed by #228

PR #203, opened 2019-03-06
For2py refactoring

adarshp:

1. I/O in `genPGM.py` has been refactored - now the string corresponding to the generated `lambdas.py` file is built in memory before being written to file - I think this will be faster and lead to easier debugging.
2. for2py modules have been moved up from the `scripts` directory. (See #164)
3. Removes the `autoTranslate` script (Obviating the need for some of the changes in #164).
4. Renames the following modules:
  - `for2py_arrays` -> `arrays`
  - `fortran_format` -> `format`
  - `f2py_pp`-> `preprocessor`
  - `fortran_runner` -> `runner`
  - `fortran_syntax` -> `syntax`
  This, along with change no. 2, leads to cleaner imports, with less redundancy - instead of writing 
  ```python
  from delphi.translators.for2py.scripts.f2py_pp import *
  ```
  we would write
  ```python
  from delphi.translators.for2py.preprocessor import *
  ```
5. Encapsulates the global variable `programName`, fixing the bug in #201 .
6. Removes residual tests and data in the `for2py` directory.
7. Updates tests, data, and docstrings.


pratikbhd:
@adarshp Referring to the discussion from Slack, here are my views on the changes you mentioned:

1. The renamings look good to me. They certainly look cleaner so I agree with this.
2. I am okay with deleting he `tests` directory under `for2py`.
3. I actually use the `autoTranslate` scripts regularly to run tests while I am making changes, but I would like to know more about the custom scripts that you are using in replacement to `autoTranslate`. For now, I would prefer keeping the `autoTranslate` script.

adarshp:
Sounds good - I reverted the deletion of `autoTranslate`, and will merge now.

PR #204, opened 2019-03-07
Ctypes

hlim1:
This PR contains the refactoring of how variable, arrays, and derived type declaration from List to ctypes. Now, the translated python output imports ctype's c_int, cfloat, and c_wchar_p. However, there are a bit of more works to be done for the float value precision, and since the branch for precision was already generated, I will separately work on this issue in that branch and PR from there.


hlim1:
Closing this PR.

Issue #200, opened 2019-03-04
Source ordering mismatch in JSON/Lambdas


pauldhein:
Sorry @pratikbhd it seems that reconciling the JSON and Lambdas for any GrFN is quickly turning into a [hydra](https://www.google.com/search?q=hydra&client=firefox-b-1-ab&source=lnms&tbm=isch&sa=X&ved=0ahUKEwi9tqad0OngAhWP2YMKHQvlD7MQ_AUIDigB#imgrc=pr5bmHiMzPdIXM:), solve one issue and several others take it's place.

Now that the variable indexing is done appropriately I have noticed that the ordering of arguments in the sources list of functions in the JSON is not in sync with the arguments as they appear in the lambda function calls. Here is an example of the problem I am seeing:

`JSON`: (function entry)
```json
{
  "name": "update_est__decision__yield_est_0",
  "type": "decision",
  "target": "yield_est",
  "reference": 11,
  "sources": [
    {
      "name": "IF_1_0",
      "type": "variable"
    },
    {
      "name": "yield_est_2",
      "type": "variable"
    },
    {
      "name": "yield_est_1",
      "type": "variable"
    }
  ]
},
```

`JSON`: (location in body of `update_est`)
```json
{
  "name": "update_est__decision__yield_est_0",
  "output": {
    "variable": "yield_est",
    "index": 3
  },
  "input": [
    {
      "variable": "IF_1",
      "index": 0
    },
    {
      "variable": "yield_est",
      "index": 2
    },
    {
      "variable": "yield_est",
      "index": 1
    }
  ]
}
```
`Lambdas`:
```python
def update_est__decision__yield_est_0(yield_est_2, yield_est_1, IF_1_0):
    return yield_est_1 if IF_1_0 else yield_est_2
```

As you can see the ordering in `input` and `sources` are are inconsistent with the ordering of the arguments in the actual lambda function. Now I could solve this by hashing the args, performing a lambda function inspection and then doing proper variable association ... but then I'd have to do that for every lambda function, and that amount of processing is non-trivial and unnecessary for us to get our GrFN representation working. Can you please update the code in `for2py` to ensure that whatever ordering of variables is used in the lambda functions is also used in **all** locations in the JSON?

pratikbhd:
[...]

PR #205, opened 2019-03-08
Handling proper variable indexing

pratikbhd:
This PR implements the following:

1. Handles the indexing of variables as discussed in #185 and #200


delphi/translators/for2py/genPGM.py
```diff
@@ -23,6 +23,10 @@
     ast.LtE: operator.le,
 }
 
+has_elif = False
```
adarshp: @pratikbhd Can you make all global variables uppercase?

adarshp:
@pauldhein Please squash and merge once you approve. Thanks!

Issue #209, opened 2019-03-10
Strange occurrence of list in PGM JSON


pauldhein:
After merging #208 it seems that one of the assignment functions for `PETPT` is being stored as a list instead of an object. Here is a view of the function in question (generated via `autotranslate`):

<img width="508" alt="Screen Shot 2019-03-10 at 10 13 36 AM" src="https://user-images.githubusercontent.com/14334973/54086206-bae20d00-431d-11e9-8ca3-d4ac8331c18f.png">

@pratikbhd Can you figure out what this is happening?

pratikbhd:
I seemed to have missed this happening when pushing the last code, I'm sorry about that. I have fixed the issue and created a pull request in #210

pauldhein:
Closed by #210

PR #210, opened 2019-03-10
Bugfix for Issue 209

pratikbhd:

1. Fix the accidental storing of assignment functions in GrFN file as a list as mentioned in #209


adarshp:
Closes #209

PR #216, opened 2019-03-14
Refactor frontend

skdebray:
I recalibrated my ambitions after the discussion with Adarsh this morning, so the amount of actual refactoring in this PR is minimal: just a couple of places where different methods with essentially identical functionality could be merged, thereby reducing code duplication.  Also, added some code to keep track of, and report warnings on, tags from the xml AST that are not explicitly handled by the xml-to-JSON translator.


adarshp:
This looks good, thanks! I pushed a couple of minor edits - once tests pass, I'll merge.

Issue #211, opened 2019-03-11
Standardize JSON handling of containers and loop plates


pauldhein:
Currently we have two different styles for handing inputs for `container`s and `loop_plate`s in our JSON.

Containers:
```json
{
  "name": "update_est",
  "type": "container",
  "input": [
    {
      "name": "rain",
      "domain": "real"
    },
    {
      "name": "total_rain",
      "domain": "real"
    },
    {
      "name": "yield_est",
      "domain": "real"
    }
  ],
}
```

Loop plate
```json
{
  "name": "crop_yield__loop_plate__day_0",
  "type": "loop_plate",
  "input": [
    "consistency",
    "max_rain",
    "absorption",
    "rain",
    "total_rain",
    "yield_est"
  ],
}
```

1. Please change the style of the input list for `loop_plate`s to match that of a `container`s
2. `rain` is not defined before the call to`crop_yield__loop_plate__day_0` so it should not be an input variable. Can we please make this change?

The two items are also handled differently when they are being called:

A call for a container:
```json
{
  "function": "update_est",
  "output": {},
  "input": [
    {
      "variable": "rain",
      "index": 0
    },
    {
      "variable": "total_rain",
      "index": -1
    },
    {
      "variable": "yield_est",
      "index": -1
    }
  ]
},
```

A call for a loop plate:
```json
{
  "name": "crop_yield__loop_plate__day_0",
  "inputs": [
    "consistency",
    "max_rain",
    "absorption",
    "rain",
    "total_rain",
    "yield_est"
  ],
  "output": {}
},
```
3. Please change the `inputs` argument for a `loop_plate` to be named `input` and change the input list to match the input list schema for a `container`
4. Once again please remove `rain` as an input

Variable indexing for inputs is currently being handled differently:

Input variable indexing in a container:
```json
"body": [
  {
    "name": "update_est__assign__total_rain_1",
    "output": {
      "variable": "total_rain",
      "index": 1
    },
    "input": [
      {
        "variable": "total_rain",
        "index": 0
      },
      {
        "variable": "rain",
        "index": 0
      }
    ]
  },
]
```

Input variable indexing in a loop plate
```json
"body": [
  {
    "name": "crop_yield__assign__rain_0",
    "output": {
      "variable": "rain",
      "index": 0
    },
    "input": [
      {
        "variable": "day",
        "index": -1
      },
      {
        "variable": "consistency",
        "index": -1
      },
      {
        "variable": "max_rain",
        "index": -1
[...]

PR #221, opened 2019-03-20
Issue 211 - points 1 and 3

adarshp:
This PR addresses points 1 and 3 of #211 .


tests/data/crop_yield.json
```diff
@@ -433,13 +450,31 @@
...
         {
           "name": "crop_yield__loop_plate__day_0",
-          "inputs": [
-            "consistency",
-            "max_rain",
-            "absorption",
-            "rain",
-            "total_rain",
-            "yield_est"
+          "input": [
```
pauldhein: I wasn't looking for index information in the input list. I am sorry if I didn't make that clear. The objects hear should just have the name and domain information.

tests/data/PETPT.json
```diff
@@ -207,17 +207,17 @@
...
         {
-          "name": "eo_2",
+          "name": "eo_3",
           "type": "variable"
         }
       ]
     },
     {
-      "name": "petpt__assign__eo_5",
+      "name": "petpt__assign__eo_6",
```
pauldhein: It seems like we have introduced an indexing bug that is indexing all variables too quickly. We should only have `eo_0` through `eo_5` represented here, but instead we are seeing `eo_6`.

delphi/translators/for2py/genPGM.py
```diff
@@ -222,13 +222,29 @@ def genPgm(self, node, state, fnNames, call_source):
...
                 "body": loopBody,
             }
 
-            loopCall = {"name": loopName, "inputs": variables, "output": {}}
+            loopCall = {
+                "name": loopName,
+                "input": [
+                    {
+                      "name": variable,
+                      "index": -1,
```
pauldhein: This line can probably just be changed to `"domain": state.varTypes[variable]` for now

tests/data/PETPT.json
adarshp: @pratikbhd Looks too hard for me to fix - I'll leave it for you and close this PR :)

pauldhein:
@adarshp as noted in #211, I messed up in reviewing your changes (I got confused by the git diff view instead of the actual file view). This PR should be re-opened and merged with master (after solving the off-by-1 indexing bug that was introduced to PETPT), but the input list for the loop plate call in the body of the crop yield container is formed correctly.

pratikbhd:
> @adarshp as noted in #211, I messed up in reviewing your changes (I got confused by the git diff view instead of the actual file view). This PR should be re-opened and merged with master (after solving the off-by-1 indexing bug that was introduced to PETPT), but the input list for the loop plate call in the body of the crop yield container is formed correctly.

Would it be alright if I implemented this change in the branch that I am working on right now regarding the indexing issue for #211? That would make the merges more streamlined.

pauldhein:
@pratikbhd that sounds like a good idea to me. BTW, any idea why the indexing error started to occur in `PETPT`? It doesn't seem to me like @adarshp changed anything, but yet the `PETPT` `EO` indices seem to start at 1 instead of 0.

PR #223, opened 2019-03-20
Function array issue

hlim1:
Fixed a python output bug when handling a function call and array syntax due to the AST ambiguity.

### DIFF
--- a/delphi/translators/for2py/pyTranslate.py
+++ b/delphi/translators/for2py/pyTranslate.py
@@ -264,7 +264,6 @@ class PythonCodeGenerator(object):
             )
         else:
             argSize = len(node["args"])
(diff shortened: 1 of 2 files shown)


delphi/translators/for2py/pyTranslate.py
```diff
@@ -263,7 +263,7 @@ def printCall(self, node: Dict[str, str], printState: PrintState):
             )
         else:
             argSize = len(node["args"])
-            assert argSize >= 1
+            # assert argSize >= 1
```
adarshp: Can you delete this line?

adarshp: Actually never mind, I can do it.

PR #224, opened 2019-03-20
For2Py custom exception class

adarshp:
This PR adds a custom exception class for for2py, `For2PyError`.

So instead of writing

```python
sys.stderr.write("Zero-length arrays current not handled!\n")
sys.exit(1)
```

you would now write

```python
raise For2PyError("Zero-length arrays current not handled!")
```

This exception will then be propagated to the CodeExplorer webapp.

To test this out, grab the latest master branches of the Delphi and automates repos and run the webapp locally (I won't be deploying this to http://vanga.sista.arizona.edu/automates until issue #211 is resolved - there are still some wiring issues).

Here's a preview of how the errors will appear:

Screen Shot 2019-03-20 at 3 21 00 PM

This will work for error messages up to 4000 characters long. If you would like longer error messages, let me know and I can implement that.

### DIFF
--- a/delphi/translators/for2py/__init__.py
+++ b/delphi/translators/for2py/__init__.py
@@ -1,0 +1,15 @@
+class For2PyError(Exception):
+    status_code=400
+
+    def __init__(self, message, status_code=None, payload=None):
+        Exception.__init__(self)
+
+        self.message = message
+        if status_code is not None:
+            self.status_code = status_code
+        self.payload = payload
+
+    def to_dict(self):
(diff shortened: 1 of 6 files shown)


adarshp:
Also FYI, we can have the error message be constructed from a more complex object than a string - for example, a dict with keys like line number, filename, etc. - this would be pretty easy to implement so let me know if that's something you need @pratikbhd @skdebray @hlim1 .

skdebray:
:+1:

PR #217, opened 2019-03-15
GroundedFunctionNetwork creation and analysis

pauldhein:
## Changes
- Implemented a `GroundedFunctionNetwork` class that can be constructed from Fortran and serves as a real exectuable representation of a `GrFN`.
- Implemented a `run()` method that allows the a `GrFN` to be run with either single or batched input examples
- Implemented a `to_ProgramAnalysisGraph` method that allows us to view the variable connections in the GrFN
- Implemented visualization methods that allow us to view a full GrFN
- Changed the implementation of Sobol analysis to work with GrFN as a single function
- Implemented FAST sensitivity analysis for quickly getting `S1` and `ST` indices
- Removed large amounts of file and folder bloat from `delphi/GrFN` and `delphi/analysis`

## Requirements before merge
1. Complete testing with `crop_yield.f` (Requires #211 )
2. Complete visualization testing with a module that contains sub-modules (Requires `1`)

## Additions coming soon
- Surface plot function from GrFN evaluation
- Testing ForwardInfluenceBlanket generation from GrFN (Requires translated `PETASCE`)

PR #228, opened 2019-03-27
GrFN class, round 2

adarshp:
Ok, I think these changes are ready to be merged into master.

This PR is a continuation of #217 . In addition to the changes there, it also implements the following:

- Reorganization of the directories - there is now an apps directory under Delphi, with `CodeExplorer` and `rest_api` (formerly `icm_api`) residing in it.
- Reorganization of test data directories
- tests have been added for the wiring of the PETPT GrFN and CAG, to prevent regression bugs.

@pauldhein - I've made the following changes that you'll probably want to review.
- I've switched from using the enum to a string comparison, mainly for readability reasons, but can revert if necessary.
- I refactored the `from_dict` method of the `GroundedFunctionNetwork` class to be a bit cleaner
- I've modified the `run` method of the `GroundedFunctionNetwork` - no need to generate a sorted list of inputs anymore - Python's dict unpacking ensures that the appropriate values get passed to the right arguments.
- I've done away with the `build_call_graph` and `build_function_sets` methods in the `ForwardInfluenceBlanket` class since they seemed to be the same as the corresponding methods in `GroundedFunctionNetwork`
- I changed the `lambda` key to `lambda_fn` to avoid conflict with the Python builtin `lambda` keyword.
- I've removed `scopes.py`


adarshp:
@pauldhein - if you are ok with these changes, please squash and merge them.

adarshp:
Closes #217 .

cl4yton:
This is a lot of work, guys, many thanks!

adarshp:
This PR now also closes #164 .

adarshp:
Closed by #229

PR #230, opened 2019-03-27
Provenance from text reading added to variable nodes in CodeExplorer

adarshp:
Screen Shot 2019-03-27 at 4 26 06 PM


adarshp:
fyi @BeckySharp

BeckySharp:
yay!

PR #231, opened 2019-03-28
End to end pipeline implemented with S2 surface viz for PETPT example

adarshp:
Screen Shot 2019-03-27 at 7 21 31 PM


pauldhein:
@adarshp Instead of S2 for the z-axis label, what if we used the name of the actual output variable? (`EO` in this case)

adarshp:
Done!

Screen Shot 2019-03-27 at 7 37 38 PM

PR #237, opened 2019-04-03
Refactor pyTranslate.py

skdebray:
This PR incorporates a bunch of refactoring and code cleanup for the file `pyTranslate.py`.  The code still needs work, but I think this may be a good point to do a merge so that we can sync up this code with @hlim1's separate effort to repair the XML output from `OFP` and refactor/cleanup the file `translate.py`.  Once these two are back in sync, we can come back and work on improving the code further while also extending it with additional features.


skdebray:
@adarshp The build failures in travis for this PR come from the fact the code in this PR doesn't handle derived types.  That, in turn, is because @hlim1 is currently, in a separate branch, "rectifying" the XML AST output by OFP (there are a lot of weirdnesses in OFP's output that complicate processing, and that for derived types in particular is a holy mess).  I didn't see any point in refactoring large chunks of code for handling derived types that, because of the code restructuring in this PR, we'll end up discarding anyway.  

W.r.t. the refactoring in this PR, I can see two ways to proceed: (1) we can wait until @hlim1's rewrites have been merged in [~7-10 days] and then add code to handle the new representation for derived types; or (2) we can temporarily skip the tests involving derived types, and add them back in once @hlim1's  code gets added in.  I don't have strong feelings either way and am happy to go with whatever you prefer.

skdebray:
@adarshp On a slightly related note, I'd like to discuss how best to test the for2py code.  Right now it looks like the Python file generated by `pyTranslate.py` file is diff'd against a reference version.  This has the problem that any changes to the generated Python (e.g., additional parens, as in the output from this PR) cause tests to fail.  Can we briefly discuss this at the PA meeting tomorrow?

adarshp:
> @adarshp The build failures in travis for this PR come from the fact the code in this PR doesn't handle derived types. That, in turn, is because @hlim1 is currently, in a separate branch, "rectifying" the XML AST output by OFP (there are a lot of weirdnesses in OFP's output that complicate processing, and that for derived types in particular is a holy mess). I didn't see any point in refactoring large chunks of code for handling derived types that, because of the code restructuring in this PR, we'll end up discarding anyway.

I agree.
> 
> W.r.t. the refactoring in this PR, I can see two ways to proceed: (1) we can wait until @hlim1's rewrites have been merged in [~7-10 days] and then add code to handle the new representation for derived types; or (2) we can temporarily skip the tests involving derived types, and add them back in once @hlim1's code gets added in. I don't have strong feelings either way and am happy to go with whatever you prefer.

Let's skip the derived type tests for now - just add the following decorator on the line above the test function(s) you want to skip.

```python
@pytest.mark.skip
```


> @adarshp On a slightly related note, I'd like to discuss how best to test the for2py code. Right now it looks like the Python file generated by `pyTranslate.py` file is diff'd against a reference version. This has the problem that any changes to the generated Python (e.g., additional parens, as in the output from this PR) cause tests to fail. Can we briefly discuss this at the PA meeting tomorrow?

Yes, let's discuss it at the PA meeting - I have a couple of thoughts on how to make refactoring less painful in general (basically, I think we should be testing the downstream GrFN wiring and execution rather than the intermediate output Python source code and GrFN JSONs).

PR #240, opened 2019-04-18
Comments

skdebray:
This PR incorporates code to automatically insert markers for internal comments.  Right now the internal comments are collected but then nothing is done with them.  Once we figure out the API for interacting with other components of the system, this code will be extended to write out the comments in an appropriate form.  At that time, the file `get_comments.py` will go away.


skdebray:
Hmm.  I got rid of the function `process()` in `preprocessor.py` and that seems to be causing a lot of tests to fail in travis.  We need to discuss how best to address this.

PR #245, opened 2019-05-24
Preprocessor to lower

skdebray:
This PR has the preprocessor convert the Fortran source code into lower-case before writing it out.  This frees us from having to constantly convert to lower-case in translate.py and pyTranslate.py.

### DIFF
--- a/delphi/translators/for2py/format.py
+++ b/delphi/translators/for2py/format.py
@@ -369,20 +369,20 @@ class Format:
 #           and-reference-rules-for-list-directed-sequential-write-statements
 
 LIST_WRITE_DEFAULTS = {
-    "BYTE": "I5",
-    "LOGICAL(1)": "L2",
-    "LOGICAL(2)": "L2",
-    "LOGICAL(4)": "L2",
-    "LOGICAL(8)": "L2",
-    "INTEGER": "I12",
-    "INTEGER(1)": "I5",
-    "INTEGER(2)": "I7",
-    "INTEGER(4)": "I12",
-    "INTEGER(8)": "I22",
-    "REAL": "1PE14.5E2",
-    "REAL(4)": "1PG15.7E2",
-    "REAL(8)": "1PG24.15E3",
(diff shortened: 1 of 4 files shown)


skdebray:
@adarshp It looks like the  checks are failing on this PR with the following error:
```
self = <delphi.translators.for2py.pyTranslate.PythonCodeGenerator object at 0x2b96e8a7a3c8>
node = {'name': 'MSALB', 'tag': 'arg', 'type': 'REAL'}
printState = <delphi.translators.for2py.pyTranslate.PrintState object at 0x2b96e8a8a390>
    def printArg(self, node, printState: PrintState):
        try:
>           var_type = TYPE_MAP[node["type"]]
E           KeyError: 'REAL'
delphi/translators/for2py/pyTranslate.py:533: KeyError
```

I looked at the preprocessed Fortran source output by the preprocessor, as well as the xml files obtained from it, and they are all in lower-case so I don't see where the upper-case string "REAL" could be coming from.  It so happens that the source file `delphi/tests/data/program_analysis/PETPT.for` (which is the test that failed) has it in upper case, but the preprocessor should have converted everything to lower case anyway.  Any thoughts?

adarshp:
The function that is used for integration with the rest of Delphi is `preprocessor.process`: 

https://github.com/ml4ai/delphi/blob/be8536c9706411ec8745b62f888be55dd4a29281/delphi/translators/for2py/preprocessor.py#L313-L324

However, the lowering happens outside of that function, in the `main` block:

https://github.com/ml4ai/delphi/blob/be8536c9706411ec8745b62f888be55dd4a29281/delphi/translators/for2py/preprocessor.py#L344

So the lowercasing never actually happens unless preprocessor.py is invoked as a script....

adarshp:
i.e. to fix it, you would need to move the lowercasing functionality inside the `process` function (and in general, for integration, it's best to keep as much of the logic within functions and outside of the main block!)

skdebray:
Ah, I get it.  Thank you. :+1:

PR #250, opened 2019-06-01
Issue 249

hlim1:
Handled the issue of function AST does not get printed by rectify.py.
Handled the issue of pyTranslate.py does not print the function properly.
Tested with the given example.

`    integer function foo(x)
      implicit none
      integer x
      foo = x + 1234567      ! return value of the function
      end function foo

      program main
      integer u, foo
      u = foo(3)
 10   format(I8)
      write (*,10) u
      end program main`

### DIFF
--- a/delphi/translators/for2py/pyTranslate.py
+++ b/delphi/translators/for2py/pyTranslate.py
@@ -171,25 +171,40 @@ class PrintState:
 
 class PythonCodeGenerator(object):
     def __init__(self):
+        # Variable to hold the program name
         self.programName = ""
+        # Dictionary to hold the tag mapping to the
+        # print function that needs to be invoked
         self.printFn = {}
+        # Dictionary to hold the declared variable
+        # and its declared type
         self.variableMap = {}
+        # List to hold the imports in the program
         self.imports = []
-        # This list contains the private functions
+        # List to hold the private functions
         self.privFunctions = []
-        # This dictionary contains the mapping of symbol names to pythonic
-        # names
+        # Dictionary to hold declared symbols (variable,
+        # function, and format, etc) as key and map to
+        # the symbol name that will be used in the python IR output
         self.nameMapper = {}
-        # Dictionary to hold functions and its arguments
+        # List to hold the declared function names
         self.functions = []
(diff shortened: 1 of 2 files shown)


delphi/translators/for2py/pyTranslate.py
```diff
@@ -703,7 +706,28 @@ def printFuncReturn(self, node, printState: PrintState):
...
+                if "value" in node:
+                    val = node["value"]
+                else:
+                    assert (
+                        "left" in node and
+                        "operator" in node and
+                        "right" in node
+                    ), f"Something is missing. Detail of node content: {node}"
+
+                    left = ""
```
skdebray: This assignment seems unnecessary, since the variable is overwritten in the if statement.

delphi/translators/for2py/pyTranslate.py
```diff
@@ -703,7 +706,28 @@ def printFuncReturn(self, node, printState: PrintState):
...
+                    if node["left"][0]["tag"] == "ref":
+                        left = node["left"][0]["name"]
+                    else:
+                        left = node["left"][0]["value"]
+                    operator = node["operator"]
+                    right = ""
```
skdebray: This assignment seems unnecessary, since the variable is overwritten in the if statement.

delphi/translators/for2py/pyTranslate.py
```diff
@@ -181,11 +181,13 @@ def __init__(self):
         # names
         self.nameMapper = {}
         # Dictionary to hold functions and its arguments
+        self.functions = []
```
skdebray: Please add a comment to each of these fields saying what it's for.  Also, please elaborate on the comment before `nameMapper`, which currently is not very helpful.

hlim1:
@skdebray As requested, I've removed the unnecessary left and right variables. It was my habit of declaring before use from C/C++, which is not needed in Python. I should look into other codes whether I'm doing the same thing, and remove them as necessary. Also, I added comments for all declared member variables, lists, and dictionaries. I hope they make more sense now.
Thank you.

PR #251, opened 2019-06-01
Floating precision fix

pratikbhd:
1. Implements the fix for floating point precision issue.


skdebray:
Please fix the build errors reported by travis.

delphi/translators/for2py/floatNumpy.py
```diff
@@ -0,0 +1,149 @@
+from numpy import float32
+
+################################################################################
+class FloatNumpy:
+    def __init__(self, val):
+        self._val = float32(val)
+
+    def __add__(self, other):
+        if isinstance(other, FloatNumpy):
```
skdebray: This if statement is repeated in every method of this class.  Such repetition of code is undesirable.  Instead, I suggest defining a method like the following:
```
def get_value(self, other):
    if isinstance(other, Float32):
        return other._val
    else:
        return other
```
Now you can define `__add__()` as:
```
def __add__(self, other):
    return self._val + self.get_value(other)
```
The other operators are analogous.

More generally, we should try to avoid repeating code.  That's bad style.  (I know you inherited this code from me, so the bad style here is originally my fault.  But it still needs to be fixed.)

delphi/translators/for2py/floatNumpy.py
```diff
@@ -0,0 +1,149 @@
+from numpy import float32
```
skdebray: This file desperately needs comments indicating its purpose.  It's entirely possible that other people may read this code.

delphi/translators/for2py/floatNumpy.py
```diff
@@ -0,0 +1,149 @@
...
+
+################################################################################
```
skdebray: To be consistent with the coding style used in other files in this system, this line (and a similar line later in this file) should be deleted.

delphi/translators/for2py/floatNumpy.py
```diff
@@ -0,0 +1,149 @@
...
+            n = other
+
+        return self._val > n
+
+    def __str__(self):
+        return str(self._val)
+
+################################################################################
+
+def eps_native():
```
skdebray: This code should not be here.  Please remove irrelevant code from this file.

delphi/translators/for2py/genPGM.py
```diff
@@ -1629,7 +1638,7 @@ def create_pgm_dict(
     """ Create a Python dict representing the PGM, with additional metadata for
     JSON output. """
 
-    lambdaStrings = ["import math\n\n"]
+    lambdaStrings = ["import math\n", "from delphi.translators.for2py.floatNumpy import FloatNumpy\n\n"]
```
skdebray: Python lines should be less than 80 characters long.

delphi/translators/for2py/pyTranslate.py
```diff
@@ -133,6 +133,8 @@ def __init__(
         functionScope="",
         indexRef=True,
         varTypes={},
+        isFloatNumpy=False,
```
skdebray: Add comments indicating the purpose of these fields.

delphi/translators/for2py/pyTranslate.py
```diff
@@ -315,23 +323,44 @@ def proc_intrinsic(self, node):
 
         arg_list = self.get_arg_list(node)
         arg_strs = [
-            self.proc_expr(arg_list[i], False) for i in range(len(arg_list))
+            self.proc_expr(arg_list[i], False, printState) for i in range(len(arg_list))
```
skdebray: line is too long.

delphi/translators/for2py/pyTranslate.py
```diff
@@ -315,23 +323,44 @@ def proc_intrinsic(self, node):
...
+        elif arg["tag"] == "op":
+            for item in arg["left"]:
+                st = self.check_ref(index, item, arg_strs)
+            if not st:
+                for item in arg["right"]:
+                    st = self.check_ref(index, item, arg_strs)
+            return st
+        elif arg["tag"] == "call":
+            for item in arg["args"]:
+                st = self.check_ref(index, item, arg_strs)
```
skdebray: add an "else" at the end of this if-elif sequence to catch anything not covered:
```
if arg["tag"] == "ref":
    ...
elif ...
   ...
elif ...
    ...
else:
    assert False
```

delphi/translators/for2py/pyTranslate.py
```diff
@@ -633,22 +666,25 @@ def printIf(self, node, printState: PrintState):
...
+        ref_str = self.proc_ref(node, False, printState)
         self.pyStrings.append(ref_str)
 
     def printAssignment(self, node, printState: PrintState):
+
         assert len(node["target"]) == 1 and len(node["value"]) == 1
         lhs, rhs = node["target"][0], node["value"][0]
 
-        rhs_str = self.proc_expr(node["value"][0], False)
+        printState.isAssignment = True
```
skdebray: I'm trying to understand why `printState` has been added as an argument to `proc_expr()`, `proc_ref()`, etc.  It has the effect of weaving the entire state of this computation through the entire computation, and makes it harder to understand and maintain the code.  I would like to avoid this if at all possible.

Is `printState` being used just to indicate whether an expression is being used on the RHS of an assignment via the `isAssignment` flag?  If so, it seems simpler to just pass down a boolean argument indicating this.

delphi/translators/for2py/pyTranslate.py
```diff
@@ -675,12 +711,17 @@ def printAssignment(self, node, printState: PrintState):
                 # target is a scalar variable
                 assg_str = f"{lhs['name']}[0]"
 
+        # Check if the lhs is a real and convert the variable to a numpy float object if it is
+        if self.variableMap.get(lhs["name"]) == "REAL" and rhs["tag"] == "literal":
```
skdebray: Please check your lines and fix those that are >= 80 characters long.

delphi/translators/for2py/pyTranslate.py
```diff
@@ -1127,24 +1169,24 @@ def get_range(self, node):
...
-            loBound = self.proc_literal(low[0])
+            loBound = self.proc_literal(low[0], printState)
         elif low[0]["tag"] == "op":
-            loBound = self.proc_op(low[0])
+            loBound = self.proc_op(low[0], printState)
         else:
             assert False, f"Unrecognized tag in upper bound: {low[0]['tag']}"
 
-        # Get upper bound value
+        # Get upper bound valuef
```
skdebray: why valuef ?

delphi/translators/for2py/pyTranslate.py
```diff
@@ -1127,24 +1169,24 @@ def get_range(self, node):
         up = node["high"]
         # Get lower bound value
         if low[0]["tag"] == "literal":
-            loBound = self.proc_literal(low[0])
+            loBound = self.proc_literal(low[0], printState)
```
skdebray: Elsewhere in this file we use underscores in variable names.  To be consistent, this should be `lo_bound` rather than `loBound`.

skdebray:
@pratikbhd Looks like interfacing with external libraries is a continuing issue.  Stop by if you'd like to discuss how we might want to address this problem.

delphi/translators/for2py/floatNumpy.py
adarshp: Could you also please set a maximum line length of 80?

delphi/translators/for2py/floatNumpy.py
```diff
@@ -0,0 +1,210 @@
...
+    arithmetic and logical operations. All methods below follow a similar input/output type.
+
+    Input: Either a Float32 variable or a variable of a supported type.
+    Output: A Float32 object of the result of the operation.
+    """
+
+    def __init__(self, val):
+        """
+        This constructor converts the float variable 'val' into numpy's float32 type.
+        :param val: float variable
```
adarshp: Please change this docstring to follow the Google docstring convention that we use in the rest of the codebase - here are some examples: https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html#example-google

pauldhein:
@pratikbhd I have reviewed the code in this PR and made some changes that I think will both solve the issues we are having and will allow for the code generated in the lambdas file to maintain it's generality (i.e. I do not want any unnecessary type dependency to be present in the lambdas).

## My Changes
- Changed `Float32` to be a subclass of Python's builtin `Real` class. This ensures that we respect the Python Number hierarchy, and allows us to use the hierarchy for type identification.
- Added `Number` hierarchy checking to the `Float32.__value` function to ensure the second argument is a numeric type (and throw an error otherwise)
- Updated `networks.GroundedFunctionNetwork.sobol_analysis` to work with Float32 and to ensure proper conversion for string arguments.
- Updated `tests/test_GrFN.py` and `tests/test_analysis.py` to test 32-bit floating point values against the returns from NumPYs float32 class.
- Created `math_ext.py` a python file that holds a collection of functions that operate over Python's base math functions. 


## Purpose of the `math_ext` file
The purpose of the `math_ext` file is to collect a set of functions that mask the functions present in Python's `math` library. This allows us to ensure that we get the proper precision in and out of math functions without having to have any of the code machinery present in the lambdas files. Consider the example lambda function below from PETASCE:
```python
def petasce__assign__ws_0(xlat: Float32, pie: Float32, ldelta: Float32):
    return Float32(math.acos(-(((1.0*Float32(math.tan(((xlat*pie)/180.0)._val)))*Float32(math.tan(ldelta._val)))._val)))
```
as compared to the alternative version that uses masked math functions to ensure proper data types while leaving the lambdas as type-general as possible and free from source code specific clutter.
```python
# imports at the top of the lambdas file
from numbers import Real
import delphi.translators.for2py.math_ext as math

# Much shorter and data type independent lambda
# NOTE: still requires Real(s), but that is the correct level of generality
def petasce__assign__ws_0(xlat: Real, pie: Real, ldelta: Real):
    return math.acos(-(((1.0*math.tan(((xlat*pie)/180.0))*math.tan(ldelta)))))


# Portion stored in math_ext (hidden from lambdas)
import math
from floatNumpy import Float32

def apply_op(num, op):
    if isinstance(num, Float32):
        return Float32(op(num._val))
    else:
        return op(num)

def tan(num):
    return apply_op(num, math.tan)

def acos(num):
    return apply_op(num, math.acos)
```
The second version of the lambda is:
1. much easier to read
2. easier for sympy to read and convert into an equation
3. maintains the approach that alternative data types (such as a regular float or a tensor of float32) could be used interchangeably with the lambdas

## Requested Changes
@pratikbhd and @skdebray I would like to suggest the following changes, and perhaps we can discuss them here.
- The lambdas generation code needs to be rewritten to produce lambdas in the style shown in the section above
- Can we rename the file from `floatNumpy` to something more informative? If this file is only going to house the `Float32` class then perhaps it should be named `Float32.py`. Alternatively, we may want to use this one file to hold many extended types we may need in the future. Perhaps then a better name for this class would be `types_ext.py`?
- We are starting to have a lot of files at the root of `for2py`. Perhaps all of the files associated with masking types/functions in Python could be put into a single folder under `for2py`? Maybe the folder could be named `py_ext/` or `py_mask/`?

skdebray:
@pauldhein Your suggestions make a lot of sense.  One thought:

> We are starting to have a lot of files at the root of for2py. Perhaps all of the files associated with masking types/functions in Python could be put into a single folder under for2py? Maybe the folder could be named py_ext/ or py_mask/?

There are two kinds of files under the `for2py` directory: those involved with the translation of Fortran code to Python/GrFN, e.g., `rectify.py` and `translate.py`; and those involved with runtime support for the translated code, e.g., `arrays.py` and `format.py`.  It would reduce clutter if we separated them into two different directories: one for translation and one for runtime support.  @pauldhein @pratikbhd @adarshp @hlim1 What do you think?

skdebray:
@pratikbhd I've listed below my thoughts regarding @pauldhein's suggestions for changes.  There are a number of items in your TODO queue, my suggestion is: finish up `SAVE`, then focus on these changes so that we can be done with the floating point precision issue, then fix the problem with handling modules.  Please holler if you have questions.

> The lambdas generation code needs to be rewritten to produce lambdas in the style shown in the section above.

I agree with @pauldhein that the second version of lambdas he described is much easier to read.  Let's go for that.

> Can we rename the file from floatNumpy to something more informative? If this file is only going to house the Float32 class then perhaps it should be named Float32.py. Alternatively, we may want to use this one file to hold many extended types we may need in the future. Perhaps then a better name for this class would be types_ext.py?

Let's go with `types_ext.py`.  It's entirely possible that, down the road, we may have to build our own support for other Fortran types, e.g., COMPLEX.

adarshp:
> @pauldhein Your suggestions make a lot of sense. One thought:
> 
> > We are starting to have a lot of files at the root of for2py. Perhaps all of the files associated with masking types/functions in Python could be put into a single folder under for2py? Maybe the folder could be named py_ext/ or py_mask/?
> 
> There are two kinds of files under the `for2py` directory: those involved with the translation of Fortran code to Python/GrFN, e.g., `rectify.py` and `translate.py`; and those involved with runtime support for the translated code, e.g., `arrays.py` and `format.py`. It would reduce clutter if we separated them into two different directories: one for translation and one for runtime support. @pauldhein @pratikbhd @adarshp @hlim1 What do you think?

Sounds good to me!

hlim1:
> @pauldhein Your suggestions make a lot of sense. One thought:
> 
> > We are starting to have a lot of files at the root of for2py. Perhaps all of the files associated with masking types/functions in Python could be put into a single folder under for2py? Maybe the folder could be named py_ext/ or py_mask/?
> 
> There are two kinds of files under the `for2py` directory: those involved with the translation of Fortran code to Python/GrFN, e.g., `rectify.py` and `translate.py`; and those involved with runtime support for the translated code, e.g., `arrays.py` and `format.py`. It would reduce clutter if we separated them into two different directories: one for translation and one for runtime support. @pauldhein @pratikbhd @adarshp @hlim1 What do you think?

I think this is a very good idea. This way, it'll be much nicer to organize the files by their belonging categories.

pratikbhd:
> @pauldhein Your suggestions make a lot of sense. One thought:
> 
> > We are starting to have a lot of files at the root of for2py. Perhaps all of the files associated with masking types/functions in Python could be put into a single folder under for2py? Maybe the folder could be named py_ext/ or py_mask/?
> 
> There are two kinds of files under the `for2py` directory: those involved with the translation of Fortran code to Python/GrFN, e.g., `rectify.py` and `translate.py`; and those involved with runtime support for the translated code, e.g., `arrays.py` and `format.py`. It would reduce clutter if we separated them into two different directories: one for translation and one for runtime support. @pauldhein @pratikbhd @adarshp @hlim1 What do you think?

@skdebray I like the idea of separating the codes into different directories based on their functionality. One idea for naming the folders could be `translation/` for all code that is responsible for translation from the Fortran code up to the GrFN and `ext/` (or `runtime/`) for code that provides runtime support during the translation process.

PR #256, opened 2019-06-07
SIR demo translation files

pauldhein:
@cl4yton I have translated the SIR model and left small helpful comments about what each of the included functions actually does. I included the original `DiscreteTimeSIR.jl` file as well, with some modifications so I could compare its output with the same random seed to that of my translated model in python. Unfortunately the seeds between Julia and Python do not line up, but the output has a similar form to where I feel confident saying the Python code is behaving as expected.

I have also included the "translated" SEIRD model (in quotes because not much translation was necessary). However, per my comment in that file it seems that some crucial details are missing to use that model. I **think** that the folks at GTRI changed the return style for a model to only return diffs in each of the 5 classes and that they handled the individual class updates outside of the model code, but they only time they use that model is to show the code for it in a cookbook, so I cannot be certain about how the wiring would differ from `DiscreteTimeSIR.jl`.

Let me know in the comments if you need anything more!


cl4yton:
Thanks @pauldhein !

@adarshp : Paul and I would like to keep this PR open until we've also developed the accompanying Fortran (recruiting help from the PA team) and also verifying with discussion with James F. (GTRI) that our implementation strategy is correct -- that way we can use the PR to track the conversation.

adarshp:
> Thanks @pauldhein !
> 
> @adarshp : Paul and I would like to keep this PR open until we've also developed the accompanying Fortran (recruiting help from the PA team) and also verifying with discussion with James F. (GTRI) that our implementation strategy is correct -- that way we can use the PR to track the conversation.

Sounds good!

adarshp:
Closing this PR, since it is superseded by #344 .

PR #277, opened 2019-06-20
Handling the SAVE statement

pratikbhd:
This PR contains the following changes and additions:

1. Implementation of the Fortran SAVE Statement for simple variables.
2. Fixing the true/floor division issue between Fortran and Python3.x.

**Note**: Please don't delete this branch after merge since I will continue working on this to add extend the SAVE statements for arrays and derived types.


delphi/translators/for2py/pyTranslate.py
```diff
@@ -460,9 +484,9 @@ def proc_ref(self, node, wrapper):
                 expr_str = ref_str
             else:
                 if (
-                    "is_arg" in node
-                    and node["is_arg"] == "true"
-                    or is_derived_type_ref
+                        "is_arg" in node
```
skdebray: I think expressions of the form `A and B or C` are easier to understand when they are parenthesized.  That way we can easily tell whether it's `(A and B) or C` or `A and (B or C)`.

delphi/translators/for2py/rectify.py
```diff
@@ -997,39 +1027,46 @@ def handle_tag_statement(
...
-                    
+                    self.statements_to_reconstruct_after[
+                        'stmts-follow-label'].append(current)
+
                     if (
-                        (parent.tag == "body"
-                        and parent.attrib['parent'] == "program"
-                        and "has-stop" in current.attrib)
-                        or self.goto_under_loop
+                            (parent.tag == "body"
```
skdebray: parenthesize to make the operator grouping easy to understand.

delphi/translators/for2py/rectify.py
```diff
@@ -521,73 +526,74 @@ def handle_tag_body(
...
                     else:
                         new_parent = current
                         # Reconstruction of statements
                         if (
-                            "parent" in current.attrib
-                            and (not self.goto_under_loop
-                            and current.attrib['parent'] == "program")
-                            or (self.goto_under_loop
-                            and current.attrib['parent'] == "loop")
+                                "parent" in current.attrib
```
skdebray: parenthesize to make the operator grouping easy to understand.

delphi/translators/for2py/genCode.py
```diff
@@ -226,6 +226,16 @@ def genCode(node, state):
                           genCode(node.op, state),
                           genCode(node.values[1], state))  
 
+    elif isinstance(node, ast.Attribute):
+        """
```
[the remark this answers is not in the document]
pratikbhd: Handled.

delphi/translators/for2py/genPGM.py
```diff
@@ -940,15 +940,17 @@ def genPgm(self, node, state, fnNames, call_source):
...
 
                 name = getFnName(
                     fnNames,
                     f"{state.fnName}__assign__{target['var']['variable']}",
                     target,
                 )
-                # If the index is -1, change it to the index in the fnName. This is a hack right now.
+                # If the index is -1, change it to the index in the fnName.
+                # This is a hack right now.
                 # if target["var"]["index"] < 0:
```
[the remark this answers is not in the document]
pratikbhd: This was code that is not currently needed and has been removed now.

delphi/translators/for2py/genPGM.py
```diff
@@ -1043,6 +1045,26 @@ def genPgm(self, node, state, fnNames, call_source):
 
             return pgms
 
+        # Handle Attributes
+        # Currently, this branch works as a bypass for the `feature_save`
```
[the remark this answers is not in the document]
pratikbhd: The comment has been updated to clarify the corresponding code.
```
# This is a fix on `feature_save` branch to bypass the SAVE statement
        # feature where a SAVEd variable is referenced as
        # <function_name>.<variable_name>. So the code below only returns the
        # <variable_name> which is stored under `node.attr`. The `node.id`
        # stores the <function_name> which is being ignored.
```

delphi/translators/for2py/genPGM.py
```diff
@@ -1043,6 +1045,26 @@ def genPgm(self, node, state, fnNames, call_source):
...
+        # branch to work-around the SAVE feature and its implementation.
+        elif isinstance(node, ast.Attribute):
+            lastDef = getLastDef(node.attr, state.lastDefs,
+                                 state.lastDefDefault)
+            if (
+                isinstance(node.ctx, ast.Store)
+                and state.nextDefs.get(node.attr)
+                and call_source != "annassign"
```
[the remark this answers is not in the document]
pratikbhd: When variables are defined for the first time in the translated code, they take the following form: `x: List[int] = [None]`. In this case, variable `x` has an `AnnAssign` node meaning a type annotated assignment and it is a `Store` context (`ctx`). In this case, we do not want to update the index of the variable which is the reason for this check. 
However, since SAVEd variables are now defined and assigned to `None` on the decorator function, this check is no longer needed in the `ast.Attribute` branch and has been deleted.

delphi/translators/for2py/genPGM.py
```diff
@@ -1499,11 +1522,25 @@ def getFnName(fnNames, basename, target):
 
 def getLastDef(var, lastDefs, lastDefDefault):
     index = lastDefDefault
-    if var in lastDefs:
-        index = lastDefs[var]
+
+    bypass_regex = (
```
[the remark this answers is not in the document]
pratikbhd: Updated to global variable and compiled to an re object.

delphi/translators/for2py/pyTranslate.py
pratikbhd: Relevant parenthesis has been added but since this code for derived types and written by @hlim1 , I would like him to confirm this change.

delphi/translators/for2py/rectify.py
pratikbhd: Relevant parenthesis has been added but since this code for derived types and written by @hlim1 , I would like him to confirm this change.

delphi/translators/for2py/rectify.py
pratikbhd: Relevant parenthesis has been added but since this part of this code seems to be for go-tos and written by @hlim1 , I would like him to confirm this change.

delphi/translators/for2py/rectify.py
hlim1: I believe this is already been grouped. For example, `parent.tag == "body"` is 'A', `parent.attrib['parent'] == "program"` is B, and `"has-stop" in current.attrib is C`, and `self.goto_under_loop` is D. Then, with current parenthesis, it represents the grouping of (A and B and C) or D. However, I've given a separate individual lines for the parenthesis, so they can be more visually seen.

delphi/translators/for2py/pyTranslate.py
hlim1: @pratikbhd @skdebray I've added a parenthesis to group the condition.

delphi/translators/for2py/rectify.py
hlim1: I confirm that the grouping was done appropriately.

delphi/translators/for2py/rectify.py
skdebray: @hlim1 Thanks, this is very helpful.  Can you also take a look at the boolean expression on lines 3493-3495 ?

hlim1: @skdebray I went through entire file just looking at the if-elif conditions and added grouping parenthesis for those seem to be ambiguously grouped or not having one, including line 3493-3495. However, if you find anything that I missed or you believe it can be done better, please let me know. Then, I will reflect it to the code immediately. Thank you!

PR #279, opened 2019-06-21
Align with testing

hlim1:
This PR includes the refactoring of testings.

Currently, both `make test` and `pyTest tests/test_program_analysis.py` holds the same code to translate fortran source code to GrFN. Also, autoTranslate that resides under the `for2py` is a bash script that executes programs individually by invoking the`main` functions, which creates a discrepancy in behaviours between the Travis test and local test. As a result, we had to spend unnecessary time on debugging to pass the Travis test. However, by integrating the functionalities into one file and let other test files (`test_program_analysis.py` and `network.py`) to simply call the (`fortran_to_grfn`) function of `f2grfn.py`, we can avoid those behaviour difference and code duplicates.

The usage of new `f2grfn.py` has no difference with `autoTranslate` externally except the command line (`$python f2grfn -f __filename_or_filepath__`).


skdebray:
@adarshp Since this PR touches `networks.py` I'll let you take a look over it and do the final merge if everything looks ok.

skdebray:
@hlim1 When you have a chance, can you also update the code in `delphi/tests/test_program_analysis.py`?  Thanks.

hlim1:
@skdebray How should the `delphi/tests/test_program_analysis.py` be updated from current code? I updated the `def get_python_source` to call the function from `f2grfn` to generate the Python IR and return the objects for GrFN test separately. If I am missing with something from what you told me on Wednesday about updating the `test_program_analysis.py`, please let me know, then I will make a change as soon as possible, so Adarsh can review and merge it. Thank you.

hlim1:
@skdebray In fact, I've updated the `delphi/tests/test_program_analysis.py` `make_grfn_dict` function to call `generate_grfn` function from `f2grfn.py`. If there are any more updates that need to be done, please let me know. Thank you!

skdebray:
@hlim1 This looks fine.  I think I may have been looking at the wrong branch by mistake -- sorry about that.  I'm happy with the code as it is now.

delphi/translators/for2py/f2grfn.py
```diff
@@ -0,0 +1,362 @@
+"""
+    This program is for replacement of autoTranslate bash script.
+    Instead of creating and using each file for next operation
+    like in autoTranslate bash scropt, it creates python object
+    and passes it to the next function. Thus, it works as calling
+    a multiple functions in a single program. This new autoTranslate.py
+    does not invoke main functions in each program.
+
+    In simplicity, it's a sinlg program that integrated the functionalities
```
adarshp: Typo: sinlg -> single

delphi/translators/for2py/f2grfn.py
```diff
@@ -0,0 +1,362 @@
...
+# OFP_JAR_FILES is a list of JAR files used by the Open Fortran Parser (OFP).
+OFP_JAR_FILES = [
+    "antlr-3.3-complete.jar",
+    "commons-cli-1.4.jar",
+    "OpenFortranParser-0.8.4-3.jar",
+    "OpenFortranParserXML-0.4.1.jar"
+    ]
+
+def generate_ofp_xml(preprocessed_fortran_file, ofp_file, tester_call):
+    """
```
adarshp: Can you add the meanings of the parameters in the docstring? See for example the `function_with_types_in_docstring` and `function_with_pep484_type_annotations` here: https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

adarshp: Also for the other docstrings...

hlim1: @adarshp Thank you for the feedback. I've fixed the docstrings by referencing the site you posted. Please do another review and let me know whether I have to make any further change. Thanks!

adarshp: @hlim1 Looks good! I've merged it.

PR #280, opened 2019-06-21
Parameterize unit selection

lchamp87x:
The parameterize() function in AnalysisGraph.py can now handle unit selection. The unit argument takes in a dictionary object that has the indicator variables as keys and the corresponding units as values. The get_indicator_value() function in assembly.py will throw a warning and uses its default setting for units if incorrect units are passed to its unit argument. 

The backcasting demo notebook now has an example of assigning units to a specific indicator variable.


lchamp87x:
Making a test case in test_AnalysisGraph.py to get more coverage.

Issue #278, opened 2019-06-20
Module translating to Python error


hlim1:
Handling module seems to be broken at the moment that `pyTranslate.py` either crashed or it does not produce the correct output. For example, with the `test_module_01.f`, the output python file only holds:
```
import sys
from typing import List
import math
from delphi.translators.for2py.format import *
from delphi.translators.for2py.arrays import *
from dataclasses import dataclass


x: List[int] = [1234]
```

One of the causes could be the merged functions in the refactored `pyTranslate.py` are not handling all cases that used to be handled separately in the before-refactor code. For example, in the `def nameMapping` function, it's always assuming that `args` list in the `node` only hold dictionaries of data. However, with the module, `args` list also holds a list of strings, which are the names of modules.

There are more causes of this bug and they need to be addressed to handle the module properly.

PR #294, opened 2019-06-27
Module Translation to Python Bugfix

pratikbhd:
This PR contains the following changes related to issue #278 

1. Fixed fortran-to-python translation issues with modules
2. Handling multiple format statements under the same declaration tag


delphi/translators/for2py/pyTranslate.py
```diff
@@ -1634,7 +1609,7 @@ def parse_args():
     for item in python_source_list:
         if item[2] == "module":
             try:
-                modFile = f"m_{item[1].lower()}.py"
+                modFile = f"tmp/m_{item[1].lower()}.py"
```
skdebray: Having hard-wired file or directory names is problematic (e.g., what if someone else owns the directory tmp/ and you don't have permission to access or create files in it?).  I would prefer to have the user supply such file names as arguments.

@hlim1 has had to deal with the same issue; at the time, I had not anticipated the additional need to create module-related files.  Maybe it would be enough for the user to specify just a directory name where all the various files from the for2py translation process, including modules, can be placed.  Can the two of you (@pratikbhd and @hlim1) please discuss this and come up with a proposal?

skdebray:
@pratikbhd Can you also add some module-related tests into the testing scripts?  I'd like to see delphi/tests/data/program_analysis/modules/test_module_08.f added, and maybe others that exercise other interesting aspects of modules.

delphi/translators/for2py/pyTranslate.py
hlim1: @skdebray @pratikbhd 
I've updated `f2grfn.py` to have more secure output file holder directory generation or write, which I believe can be applied to other program files (translate.py, pyTranslate.py). What I added is by either receiving the target output directory from the user as a command line `$ python f2grfn.py -f __fortran_file__ -d __directory_name__` or not and use the default directory `tmp` to generate or write to in the `parse_arg` function. Then, in the main `fortran_to_grfn` function, I'm checking whether the directory name (either it's default `tmp or given by the user) exist or not. If it does not exist, then simply create one. If it does exist, then check whether the directory is writable `os.access(temp_dir, os.W_OK)` and this is in the assert, so if it fails, the assert will give an error message the directory is not writable and terminate the program. If everything went fine without any error, then this output directory will be the place where all output files will be stored.

Therefore, this issue that you mentioned can be resolved by doing above checks in the pyTranslate.py's main and parse_arg function and use that target directory for the module file to be stored as well. For example, in the `f2grfn.py`'s `generate_python_src` function received `temp_dir` argument and passes that to generate the`modFile`'s path string. `modFile = f"{temp_dir}/m_{item[1].lower()}.py"` This way, we are not hard-wiring the directory name as well as guaranteed that the directory is writable even if the directory name was provided by the user.

skdebray:
@pratikbhd @hlim1 Thank you both for dealing with this issue so quickly.  :+1:

PR #312, opened 2019-07-16
GrFN Format Update - Incremental

pratikbhd:
This PR makes the following changes to `genPGM.py` for the GrFN specification:

1. Changed code structure to be easier to debug and scale.
2. Changed minor specification errors.
3. Renamed various variables, and methods to follow closely with the python PEP rules.

Please do not delete this branch after merging since further changes to the GrFN format will be made in this branch.


delphi/translators/for2py/f2grfn.py
```diff
@@ -473,7 +473,7 @@ def fortran_to_grfn(
         with open(original_fortran_file_path, "r") as f:
             inputLines = f.readlines()
     except IOError:
-        assert False, "Fortran file: {original_fortran_file_path} Not Found"
+        assert False, f"Fortran file: {original_fortran_file_path} Not Found"
```
skdebray: Since execution only gets to this point because an exception has been raised --- i.e., we know that things have gone wrong --- an `assert` seems not quite the right construct here.  The following seems conceptually more straightforward to me:
```
sys.stderr.write( ... )
sys.exit(1)
```

delphi/translators/for2py/genPGM.py
```diff
@@ -61,409 +67,763 @@
...
+                variable_types=local_variable_types,
+            )
+            process_decorators(node.decorator_list, function_state)
+
+        # Check if the function contains arguments or not. This determines
+        # whether the function is the outermost scope (does not contain
+        # arguments) or it is not (contains arguments). For non-outermost
+        # scopes, indexing starts from -1. For outermost scopes, indexing
+        # starts from 0
+        # TODO: What do you do when a non-outermost scope function does not
```
skdebray: @pratikbhd Is it possible to add an `assert` somewhere in for2py (not necessarily in gen_PGM.py, but wherever it makes sense) to catch this situation?  That way we'll easily know what the problem is if this situation every crops up.

delphi/translators/for2py/genPGM.py
```diff
@@ -61,409 +67,763 @@
...
+
+        pgm["functions"] += reduce(
+            (lambda x, y: x + y["functions"]), [[]] + ifPgm
+        ) + reduce((lambda x, y: x + y["functions"]), [[]] + elsePgm)
+
+        pgm["body"] += reduce(
+            (lambda x, y: x + y["body"]), [[]] + ifPgm
+        ) + reduce((lambda x, y: x + y["body"]), [[]] + elsePgm)
+
+        updatedDefs = [
```
skdebray: This list comprehension is complicated and hard to understand.  If possible, I'd like to try and follow Google's [coding style guidelines for list comprehensions](http://google.github.io/styleguide/pyguide.html#27-comprehensions--generator-expressions), esp. the "Decision" section.  Is it possible to rewrite this code to follow Google's guidelines?

delphi/translators/for2py/genPGM.py
```diff
@@ -61,409 +67,763 @@
...
+            for var in set(startDefs.keys())
+                .union(ifDefs.keys())
+                .union(elseDefs.keys())
+            if var not in startDefs
+               or ifDefs[var] != startDefs[var]
+               or elseDefs[var] != startDefs[var]
+        ]
+
+        defVersions = {
+            key: [
```
skdebray: Complicated and hard-to-understand comprehension.  See above.

PR #335, opened 2019-08-25
Bugfix

adarshp:
This PR reverts the node iterator - there is some weird behavior that makes the concept to indicator mapping not work properly.

@manujinda @lchamp87x I'm also removing the invocation of ./build/cpptests from the Makefile test target since our Python and C++ tests tend to get out of sync, and it'll be easier just to maintain the downstream ones (i.e. Python).


lchamp87x:
I discovered another bug. The prediction sequences have lost their variance, in other words all sequences have exactly the same values.

PR #338, opened 2019-08-28
Program Analysis GrFN Generation end-to-end

pratikbhd:
This PR handles the following changes:
1. Implemented the new GrFN spec.
2. End to end handling for the SIR-Gillespie code.

PS: There are some merge conflicts which will be worked on shortly. The tests will probably fail currently while we're getting the tests to pass.


skdebray:
@pauldhein I didn't realize that we need to merge this into the SIR-demo branch in order to get all the tests to pass: thanks for pointing that out.  If @adarshp is OK with that merge, I'm fine with it.

adarshp:
> @pauldhein I didn't realize that we need to merge this into the SIR-demo branch in order to get all the tests to pass: thanks for pointing that out. If @adarshp is OK with that merge, I'm fine with it.

Yes, please merge grfn_array into SIR-demo. Thanks!

PR #343, opened 2019-09-03
PA/MA Pipeline Update and GrFN2WD Translation

pauldhein:
This new branch includes all of the material from #256 that is needed to support the SIR models and collaborative demo with GTRI and Galois. The reason for this new branch is to avoid merge errors with master.


pauldhein:
This branch is tainted with the failed commits from `SIR-demo`. Closing this PR now.

PR #349, opened 2019-09-06
Source comment modification

pratikbhd:
This PR adds the following changes to the GrFN:

1. The `source_comments` field in the GrFN now contains the source comments of the original file in the form of a dictionary object instead of a string.
2. The keys in the source comments dictionary with the function names are now replaced by their respective container names.


pauldhein:
These are minor changes, and I need them immediately, so I am going to proceed with the merge.

PR #375, opened 2019-10-21
Handling loop breaks and returns

pratikbhd:
1.  Handling restructuring of breaks and returns inside loops

Note: Please do not delete this branch after merging since any updates on this will be done through this branch.


delphi/translators/for2py/loop_handle.py
```diff
@@ -0,0 +1,258 @@
...
+
+
+class RefactorBreaks(object):
+    """This class defines the refactor state of the intermediate AST while
+    removing the breaks and returns inside while loops
+    """
+
+    def __init__(self):
+        self.find_opp = {
+            ".eq.": ".ne.",
```
skdebray: This table of complementary operators also appears in rectify.py.  Can we pull this out into syntax.py?

delphi/translators/for2py/loop_handle.py
```diff
@@ -0,0 +1,258 @@
...
+        self.is_break = False
+        self.shifted_items = {}
+        self.shifted_level = 0
+
+    def refactor(self, ast: Dict) -> Dict:
+        body = ast["ast"][0]["body"]
+        self.search_while(body)
+
+        # self.search_breaks(body)
+        ast["ast"][0]["body"] = body
```
skdebray: I don't understand lines 47 and 51.

delphi/translators/for2py/loop_handle.py
pratikbhd: Just put the dictionary into `syntax.py` and updated their usage here and in `rectify.py`

delphi/translators/for2py/loop_handle.py
pratikbhd: This is redundant and not required. Removed line 51.

PR #381, opened 2019-10-28
Grfn module

hlim1:
This PR holds single- and multiple-modules that reside in a single file to be translated into own GrFN files and link with `system.json`.
- Each module will have its own `_GrFN.json` and `_lambdas.py`.
- `system.json` will have an updated version of GrFN spec that it will contain a list of dictionaries,
where each dictionary is a single `system_def`.
- There were some major changes in `f2grfn.py` and `genPGM.py` to make the pipeline working
and pytests to pass.
- `SIR-Gillespie-SD_multi_module.f` has been added for a test.


delphi/translators/for2py/f2grfn.py
```diff
@@ -217,30 +222,31 @@ def generate_python_src(
...
-                        f.write(item[0])
-                except IOError:
-                    assert False, f"Unable to write to {python_file_name}."
+    output_list = []
+    for item in python_source:
+        if item[2] == "module":
+            module_file = f"{temp_dir}/m_{item[1].lower()}.py"
+            try:
+                with open(module_file, "w") as f:
+                    output_list.append("m_" + item[1].lower() + ".py")
```
skdebray: Would it be possible to pull the mapping from a Fortran module name to the Python file name into a separate function?  I'm worried what happens if there happens to be an already-existing file whose name begins with "m_", in which case the current approach won't work.  Having the mapping done in one place (in this new function) will make the problem easier to fix down the road.

hlim1: @skdebray I've separated out the coded and put it into its own function. Though, I feel this is like a temporary solution. I will think about how the more long-term solution that we can avoid the possible name conflict.

delphi/translators/for2py/f2grfn.py
```diff
@@ -217,33 +222,50 @@ def generate_python_src(
...
+        output_list (list): A list that holds list of output files.
+        python_files (list): A list that holds the generated python
+        python file paths.
+    ReturnL
+        None
+    """
+    module_file = f"{temp_dir}/m_{item[1].lower()}.py"
+    try:
+        with open(module_file, "w") as f:
+            output_list.append("m_" + item[1].lower() + ".py")
```
skdebray: The file name construction is repeated in lines 261 and 264.  The file name should be constructed in exactly one place.

delphi/translators/for2py/f2grfn.py
```diff
@@ -268,34 +290,77 @@ def generate_grfn(
...
+    # mapper
+    file_match = re.match(filename_regex, python_filename)
+    assert file_match, "Invalid filename."
+
+    path = file_match.group("path")
+    filename = file_match.group("filename")
+
+    # Ignore all python files of modules created by `pyTranslate.py`
+    # since these module files do not contain a corresponding XML file.
+    if not filename.startswith("m_"):
```
skdebray: This code assumes that a "m_" prefix implies a module file.  As noted in a previous comment, this is a brittle assumption and we should try to minimize our reliance on it.

delphi/translators/for2py/genPGM.py
```diff
@@ -3474,18 +3498,52 @@ def create_grfn_dict(
...
+        path = file_match.group("path")
+        filename = file_match.group("filename")
+
+        # Since we do not have separate variable pickle file
+        # for m_*.py, we need to use the original program pickle
+        # file that module resides.
+        module_name = None
+        if module_file_exist:
+            module_file_path = file_name
+            module_name = filename[2:]  # First two elements are "m_"
```
skdebray: The assumption about "m_" prefix is brittle.  I would like to minimize our reliance on it.

delphi/translators/for2py/genPGM.py
```diff
@@ -3474,18 +3498,52 @@ def create_grfn_dict(
...
+    if len(generator.mode_mapper["use_mapping"]) > 0:
+        for user, module in generator.mode_mapper["use_mapping"].items():
+            if (
+                    (user in generator.module_names)
+                    or (module_name and module_name == user)
+            ):
+                module_paths = []
+                for import_mods in module:
+                    for mod_name, target in import_mods.items():
+                        module_path = path + "m_" + mod_name + "_GrFN.json"
```
skdebray: It looks like the assumption about "m_" prefix is everywhere.  Please look for them all carefully and fix them so that future changes to the naming scheme are easy to implement.

delphi/translators/for2py/f2grfn.py
hlim1: @skdebray I've reflected your suggestion in the code by declaring`MODULE_FILE_PREFIX` variable that holds` `m_` string. Reset of the code will use this variable to form the module file name or reference to check whether it's a module file or not. We can simply modify this variable in the future if we wish to specify the module file name with another prefix before the module name.

delphi/translators/for2py/genPGM.py
hlim1: @skdebray I've made the same change to `genPGM.py` like in `f2grfn.py`. All module file checking and forming will be done with `module_file_prefix` variable rather than hardcoded `"m_"` string.

PR #388, opened 2019-11-16
Implementation of String in Delphi

pratikbhd:
This PR introduces the following changes:

1. Implementation of the `character` (strings) Fortran data type in the translation process.
2. Complete handling of strings from the Fortran up to the GrFN/Lambda generation.


skdebray:
@pratikbhd Thank you for taking care of this.

PR #413, opened 2019-11-27
Grfn derived type

hlim1:
This PR holds GrFN and Lambda functions generation handler for `derived-types`.
Derived type (Python `class`) information gets populated in the `types` attribute.
It can handle a single, multiple, array, and derived-type of derived-type attributes.

This PR also holds a modification in the `test_program_analysis.py` to include:
(1) Generated lambda file comparisons, which was not tested before.
(2) Derived-type GrFN and lambda file tests.


skdebray:
@hlim1 Thanks for taking care of this.  Just so I know where things stand: can you please let me know which test inputs involving derived types we can now take all the way to GrFN?  Thanks.

hlim1:
> @hlim1 Thanks for taking care of this. Just so I know where things stand: can you please let me know which test inputs involving derived types we can now take all the way to GrFN? Thanks.

@skdebray  No problem. The test file that was added to Travis is `/delphi/tests/data/program_analysis/derived-types/derived-types-01.f`, which holds the most complex derived type structure compare to other test files. It holds a multiple derived type declaration and one uses other derived type for its attribute type. And,  accessing these attributes requires `x%y%z` structure.

skdebray:
@hlim1 I'm looking at that program and I don't see any expression with multiple field accesses of the form x % y % z ?

Does the code work for the other test files in that directory?

hlim1:
@skdebray I'm really sorry! It's test file 04, not 01. I made a typo in the previous comment. It's `/delphi/tests/data/program_analysis/derived-types/derived-types-04.f`, and, yes, the code generates GrFN and lambdas for all files following the spec.

skdebray:
@hlim1 That's great.  Thank you for the update.

OFP's problems seem to be with assignment statements that have multiple % operators on the RHS, e.g.:

```
x = p % q % r
```
This is an OFP problem and I'm not sure we can do anything about it at our end...

Strangely, it seems to be able to handle multiple % operators on the LHS of assignment statements (as in the file `derived-types-04.f`).  

Anyway, let's try out the code in `/multifile_multimod/mfmm_01/` to see how we do.

hlim1:
@skdebray Yes, I will run `f2grfn` with the programs in `/multifile_multimod/mfmm_01/`.

Issue #308, opened 2019-07-12
Issue with double asterisks write 


hlim1:
There is an issue when both of the formats are asterisks (*). For example,

`write (*,*) __output_value__`

```
      program odd_number
      implicit none
      integer :: N, k
      N = 19
      WRITE(*,*) 'Odd number between 0 and ', N
      end program odd_number
```

The output should be the standard output of the target output value.
However, currently, the crash occurs at the level of `translate.py`.

pratikbhd:
Further information about this issue. Fix for this issue is being made to the `write_bugfix` branch. The double asterisk `write` fails at two levels.

1. (**Already Fixed**) For simple variables, the intermediate Python file gets generated but fails to execute. The following error shows up in `format.py`
```
fmt = fmt.strip()
AttributeError: 'NoneType' object has no attribute 'strip'
```
This is because the `list_output_formats` function in `format.py` is sent an argument of the form `["{'type': 'INTEGER', 'parameter': False}","{'type': 'INTEGER', 'parameter': False}",]` instead of `["INTEGER","INTEGER",]`.
This can be solved by converting the following lines in `pyTranslate.py`:

- Lines `1134` and `1108`: Change
```
self.pyStrings.append(f'"{self.variableMap[var.strip()]}",')
```
to 
```
self.pyStrings.append(f'"{self.variableMap[var.strip()]["type"]}",')
```

2. (**Not Fixed Yet**) The double asterisk `write(*,*)` fails when a string is referenced instead of a variable. The example posted by @hlim1 in the comment above is a case of that. This code fails at `pyTranslate.py` with the following error:
```
line 1135, in printWrite
    f'"{self.variableMap[var.strip()]["type"]}",'
AttributeError: 'NoneType' object has no attribute 'strip'
```

pratikbhd:
@skdebray An unhandled case for the `WRITE` function was detected. Currently, we do not handle writing out (in terminal or in a file) LOGICAL constants i.e. `.TRUE.`/`.FALSE.`.
The following error shows up in `format.py` when trying to run the intermediate Python file where a LOGICAL constant is being printed.
```
line 359, in gen_output_fmt_1
f"ERROR: Unrecognized format specifier {fmt[0]}\n"
delphi.translators.for2py.For2PyError
```

adarshp:
@pratikbhd @hlim1 Is this issue resolved?

PR #422, opened 2019-12-05
Select-Case conversion for Fortran

pratikbhd:

1. Implementation of Fortran select-cases in the Python IR and subsequent GrFN representation.
2. Added test cases to test for select-case conditions (python, GrFN, and lambda file comparison test)


delphi/translators/for2py/preprocessor.py
```diff
@@ -170,13 +171,40 @@ def process_includes(lines, infile):
 
     return lines
 
+
+def refactor_select_case(lines):
+    """Search for lines that are CASE statements and refactor their structure
+    such that they are always in a i:j form. This means any CASE statement that
+    is in the form <:3> will be <Inf:3>. This is done so that the FortranOFP
```
skdebray: This seems odd.  Should <:3> be transformed to <(minus infinity):3> ?

tests/data/program_analysis/select_case/select03.f
```diff
@@ -0,0 +1,25 @@
...
+      PROGRAM MAIN
+
+      IMPLICIT NONE
+
+      INTEGER :: I = 5
+      INTEGER :: X = 40
+      INTEGER :: Y
+      INTEGER :: Z = 2
+
+      SELECT CASE(I)
```
skdebray: Would it be possible to put a loop around the select-case so that all of the various possibilities in the CASE labels get exercised?  E.g., for `CASE (:2, 5, 9:)` I'd like to have (i) a value that exercises `:2`, (ii) a value  that exercises `5`, and (iii) a value that exercises `9:`; and similarly for the other case labels.

tests/data/program_analysis/select_case/select02.f
```diff
@@ -0,0 +1,34 @@
...
+      PROGRAM MAIN
+
+      IMPLICIT NONE
+
+      INTEGER :: I = 5
+      INTEGER :: X = 40
+      INTEGER :: Y
+      INTEGER :: Z = 2
+
+      SELECT CASE(I)
```
skdebray: Would it be possible to add a loop around the `SELECT` statement so that all of the case labels are exercised?

delphi/translators/for2py/preprocessor.py
pratikbhd: That's correct. Since I was using `Inf` just as a placeholder to help me know when a `<=` or `>=` needs to be written and the preprocessed Fortran code with `Inf`s added to it would be invalid anyway, I didn't think about making it semantically correct. I'll add the `-Inf` right now.

tests/data/program_analysis/select_case/select03.f
pratikbhd: I'll make the changes.

delphi/translators/for2py/preprocessor.py
skdebray: I see.  I didn't realize that Inf was just a placeholder.  In that case, it's fine to ignore my comment about "minus infinity" -- maybe just add a comment saying that Inf is a placeholder.

tests/data/program_analysis/select_case/select03.f
pratikbhd: Modified the test file.

tests/data/program_analysis/select_case/select02.f
pratikbhd: Modified the test file.

skdebray:
@pratikbhd I was thinking about the "Inf" placeholder used to normalize CASE labels.  Suppose the program has a variable named Inf, as shown in the program below.  Would that cause a name conflict in the translation?

```
      program f
      implicit none
      integer i
      integer, parameter :: Inf = 5
      do i = 1, 10
         select case (i)
         case (:Inf)
            write (*,*) i

         case (Inf+1:)
            write (*,*) -i
         end select
      end do

      end program f
```

pratikbhd:
> @pratikbhd I was thinking about the "Inf" placeholder used to normalize CASE labels. Suppose the program has a variable named Inf, as shown in the program below. Would that cause a name conflict in the translation?
> 
> ```
>       program f
>       implicit none
>       integer i
>       integer, parameter :: Inf = 5
>       do i = 1, 10
>          select case (i)
>          case (:Inf)
>             write (*,*) i
> 
>          case (Inf+1:)
>             write (*,*) -i
>          end select
>       end do
> 
>       end program f
> ```

@skdebray I realized that the regular expression in `preprocessor.py` only caught numeric values in the `case` statement. I expanded the regular expression now to detect variables as well.

Now, the code above does not cause conflict in the translation because of the following reason:
1. During the check for `Inf` or `-Inf` in the XML, the actual `Inf` variable will have already been converted into lowercase and the check passes without conflict.

However, I found a bug with using `write(*, *)` in the Fortran code and it fails in `format.py`. I am using `write(*, 10)` for now until I fix the issue with `write(*, *)`.

All the changed code is in the `fortran_data` branch.

skdebray:
@pratikbhd Re: constants in case labels, we also need to be able to handle string and LOGICAL constants in case labels (e.g., see [here](https://pages.mtu.edu/~shene/COURSES/cs201/NOTES/chap03/select.html)).  For example, the following code is from the file `delphi/tests/data/program_analysis/ORIG-PETASCE.for`:

```
     Subroutine GET_Real(ModuleName, VarName, Value)
      ...
      SELECT CASE (ModuleName)
      Case ('SPAM')
        SELECT CASE (VarName)
        Case ('AGEFAC'); Value = SAVE_data % SPAM % AGEFAC
        Case ('PG');     Value = SAVE_data % SPAM % PG
        Case ('CEF');    Value = SAVE_data % SPAM % CEF
        ...
        Case ('EVAP');   Value = SAVE_data % SPAM % EVAP
        Case DEFAULT; ERR = .TRUE.
        END SELECT
        ...
```

pratikbhd:
@skdebray Strings are handled by CASE statements. I haven't extensively tested for them though. I also will have to test for LOGICAL constants. 
It seems like there is a bug with `WRITE(*,*)` statements right now. I'm looking into fixing the bug after which I'll do the tests for the cases discussed above.

skdebray:
@pratikbhd Let's not worry about `WRITE (*,*)` for now, I don't think that's critical for DSSAT.  I'd rather get SELECT-CASE finished and then make progress on `DATA` statements.

Can you create an issue for `WRITE (*,*)` and assign it to me?

pratikbhd:
@skdebray Apparently there is already an issue for this #308 
I have added the current issues related to this including part of the fix I made for it yesterday. These fixed are in the branch `write_bugfix`.

PR #427, opened 2020-01-27
character under derived types

hlim1:
This PR holds fixes for the bugs in `character` (`String`) variable declaration
under derived types.

- `character` variables declare with a regular `str` types. For example,
     - `character (len=2) crop` to `crop : str` not `crop = Sting(2)`
- Multiple variables declared under a single type lead to an error. For example,
     - `character (len=8) model, ename`

Updated `interface_03.f` test file to include these handled cases.

### DIFF
--- a/delphi/translators/for2py/pyTranslate.py
+++ b/delphi/translators/for2py/pyTranslate.py
@@ -33,7 +33,7 @@ from delphi.translators.for2py import For2PyError, syntax
 
 # TYPE_MAP gives the mapping from Fortran types to Python types
 TYPE_MAP = {
-    "character": "str",
+#     "character": "str",
     "double": "float",
     "float": "float",
     "int": "int",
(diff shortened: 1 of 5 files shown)


delphi/translators/for2py/pyTranslate.py
```diff
@@ -1703,10 +1706,26 @@ def get_type(self, node):
...
+            ):
+                variable_type == "str"
+                self.set_default_var_type(var_name, variable_type)
             else:
                 assert False, f"Unrecognized variable type: {variable_type}"
 
+            return variable_type
+
+    def set_default_var_type(self, var_name, var_type):
+        """ Thu=is function sets the default variable type of the declared
```
skdebray: Typo: Thu=is

delphi/translators/for2py/rectify.py
```diff
@@ -1085,6 +1088,7 @@ def handle_tag_type(
             elif (
                     child.tag == "component-decl"
                     or child.tag == "component-decl-list"
+                    or child.tag == "component-decl-list__begin"
```
skdebray: I think something like `child.tag in ["component-decl", "component-decl-list", "component-decl-list__begin"]` might be simpler?

hlim1:
@skdebray I made changes to those two files to reflect your requests.

skdebray:
@hlim1 Thank you for taking care of this. :)

PR #441, opened 2020-02-25
Aske khan

skhan1020:
sensitivity.py can. now handle all penman-monteith models. Bugs in FAST and RBD_FAST methods (minor fix) - do not merge with master yet.

visualization.py can plot sensitivity indices (first, second order) and runtime (sampling, execution, analysis).


delphi/GrFN/linking.py
```diff
@@ -74,7 +74,7 @@ def get_call_conventions(GrFN):
 
 
 def output_link_graph(G):
```
adarshp: @pauldhein suggested change: add an optional argument called `filename`, with the default value being `linking_graph.pdf`, and replace `"linking-graph2.pdf"` on line 78 with `filename`.

delphi/GrFN/utils.py
```diff
@@ -1,17 +1,19 @@
...
-    def timed(*args, **kw):
-        ts = time.time()
-        result = method(*args, **kw)
-        te = time.time()
-        print(f"{method.__name__}:\t{((te - ts) * 1000):2.4f}ms")
-        return result
-
-    return timed
+def timeit(func):
+    """PRecord the runtime of the decorated function."""
```
adarshp: @pauldhein type: PRecord -> Record

pauldhein: Done. Thanks for catching this

delphi/GrFN/linking.py
pauldhein: Good point, I've updated the code.

PR #444, opened 2020-02-26
Dssat pet

skdebray:


skdebray:
@adarshp I looked through the log but most of the output is stuff I'm unfamiliar with and I couldn't tell whether the problem that caused the checks to fail is something in the code I'm trying to merge or whether it's somewhere else.  When you have a second (no rush!) can you take a look?  Thanks.

adarshp:
@skdebray It's not related to your code. Apparently git requires an email address for git pull? But that's not set in the Docker image. I'll work on a fix now.

PR #447, opened 2020-03-02
F2grfn

hlim1:
This PR holds some fixes for librified `f2grfn` and `f2grfn_standalone.py`.
Previously, there were some issues regarding generated files. For example,
generated files were stored at random locations that some in the `delphi`
and some in the test directory. It also disabled the functionality to designating
the target directory with a command-line argument.

The fixes include (1) store generated files under one directory either in the default directory
or user-specified directory, and (2) allow the user to specify the target directory.


delphi/translators/for2py/preprocessor.py
```diff
@@ -226,11 +226,8 @@ def get_preprocessed_lines_from_file(infile):
         lines = preprocess_lines(inputLines, infile)
         return lines
 
-def create_preprocessed_file(infile, outfile=None, forModLogGen=False):
+def create_preprocessed_file(infile, forModLogGen=False):
```
adarshp: @hlim1 Since this function does not create a file anymore, could you please rename it to something else? Or even better - combine the `get_preprocessed_lines_from_file` and `create_preprocessed_file` functions into one function, since they are virtually identical.

scripts/f2grfn_standalone.py
```diff
@@ -45,7 +44,7 @@
         "--directory",
         nargs="*",
         help="A temporary directory for generated files to be stored.",
-        default=".",
+        default=["."],
```
adarshp: @hlim1 I see now why this change is necessary for the script to work - but since we are only going to specify one temporary directory, I suggest removing the 'nargs="*" line (line 45) and reverting back to `default="."`.

scripts/f2grfn_standalone.py
```diff
@@ -74,7 +73,7 @@
         processing_modules,
     ) = f2grfn.fortran_to_grfn(
         args.file,
-        args.directory,
+        args.directory[0],
```
adarshp: @hlim1 If you do the change in my previous comment, this will need to be reverted back to `args.directory,`.

delphi/translators/for2py/preprocessor.py
hlim1: @adarshp I've merged two functions and only left `get_preprocessed_lines_from_file`.

scripts/f2grfn_standalone.py
hlim1: @adarshp I've reflected your suggestion and removed `nargs="*"` and reverted back to `default="."` and `args.directory`.

PR #449, opened 2020-03-02
Adding GrFN JSON output method

pauldhein:
This PR adds the `to_json` and `to_json_file` methods to the GroundedFunctionNetwork class that enables outputting a GrFN in our new GrFN JSON format. Future changes to the format will be necessary once we decide upon the subtle changes that still need to be made to the GrFN JSON.

### DIFF
--- a/delphi/GrFN/networks.py
+++ b/delphi/GrFN/networks.py
@@ -555,8 +555,9 @@ class GroundedFunctionNetwork(ComputationalGraph):
                 nodes_json.append(
                     {
                         "name": name,
+                        "type": "variable",
                         "reference": None,
-                        "type": {
+                        "data-type": {
                             "name": "float32",
                             "domain": [("-inf", "inf")],
                         },
@@ -568,8 +569,9 @@ class GroundedFunctionNetwork(ComputationalGraph):
                 nodes_json.append(
                     {
                         "name": name,
-                        "inputs": data["func_inputs"],
+                        "type": "function",
                         "reference": None,
+                        "inputs": data["func_inputs"],
                         "lambda": source_code,
                     }
                 )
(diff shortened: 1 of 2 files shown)


delphi/GrFN/networks.py
```diff
@@ -446,7 +457,7 @@ def from_python_src(
         module_log_file_path: str,
         mod_mapper_dict: list,
         processing_modules: bool,
-        save_intermediate_files: bool = False,
+        save_file: bool = False,
```
adarshp: Looks like this argument is not being used - let's get rid of it?

delphi/GrFN/networks.py
```diff
@@ -462,20 +473,18 @@ def from_python_src(
...
+        rectified_xml_filename = "rectified_" + str(Path(python_file)).replace(
+            ".py", ".xml"
+        )
+        os.remove(rectified_xml_filename)
         return G
 
     @classmethod
     def from_fortran_file(
-        cls, fortran_file: str, tmpdir: str = ".", save_intermediate_files: bool = False
+        cls, fortran_file: str, tmpdir: str = ".", save_file: bool = False
```
adarshp: Same issue here - save_file is not used elsewhere in the code so we might want to remove it...

delphi/GrFN/networks.py
```diff
@@ -506,7 +514,7 @@ def from_fortran_file(
             module_log_file_path,
             mod_mapper_dict,
             processing_modules,
-            save_intermediate_files = save_intermediate_files
+            save_file=save_file,
```
adarshp: If save_file is removed in the other two places, it should be removed here as well.

pauldhein:
@adarshp `save_file` has been removed. Any other changes needed before merging this branch?

adarshp:
Nope! I'll merge it.

PR #462, opened 2020-04-27
Bug fixes to get_comment.py

skdebray:
Addressing comments from @pauldhein in slack: 

During the text reading meeting yesterday it became apparent that there is a bug in the part of the PA pipeline that extracts head/neck/foot source comments and associates them with the proper subroutine or file locations. The bug seems to be a miss-association when there are multiple subroutines/functions present in a single file.
In addition it seems that we need to make a choice about how to represent block comments when there is no clear distinction between a file_head block comment and the head block comment for the first subroutine. In this case, I would like to duplicate the initial block comment so that it is extracted as the block comment for the file head and the head block comment for the first subroutine. I would like to have a similar pattern of duplication done for the foot block comment of a Fortran file. Does this make sense?
I wanted to make these changes on my own, but I am having trouble with finding time to make the changes. Can someone on the PA team take some time to review this issues above and make the necessary changes?
Attached is the PETPNO.for source code that includes source comments for PETPNO , VSPAT, and VSLOP. I have also attached the output _AIR.json I get when running scripts/f2grfn_standalone.py on the Fortran file. You can verify the bug I have described above using the source_comments portion of the attached _AIR.json. If you wish to re-run the script with the same version of PETPNO.for, then please checkout the dssat_pet branch of the delphi repository.

### DIFF
--- a/delphi/translators/for2py/get_comments.py
+++ b/delphi/translators/for2py/get_comments.py
@@ -75,8 +75,8 @@ def get_comments(src_file_name: str):
     comments = OrderedDict()
     lineno = 1
 
-    comments["$file_head"] = []
-    comments["$file_foot"] = []
+    comments["$file_head"] = None
+    comments["$file_foot"] = None
 
     _, f_ext = os.path.splitext(src_file_name)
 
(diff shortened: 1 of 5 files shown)


hlim1:
@skdebray I've replaced old `_GrFN.json` files with new ones, which fixed the pyTest errors. I believe this PR is fine to be merged with the `master`.

skdebray:
@hlim1 Thank you!  That's hugely appreciated.

PR #465, opened 2020-04-29
Fix to retain empty lines in comments

skdebray:


hlim1:
@skdebray I've pushed the updated `GrFN.json` files. All tests are now passing, so I think this branch is ready to merge with `master`.

PR #470, opened 2020-09-16
Delphi fix math

manujinda:
1. Corrected the thetas not being converted to betas before using in the LDS formulas error.

2. Implemented the matrix exponential version of the transition matrix.

3. Implemented an option to choose between matrix exponential solution and discretized solution.


adarshp:
@manujinda looks good to me - go ahead and merge!

PR #473, opened 2020-10-12
Delphi update rest api 2 adarsh

manujinda:


manujinda:
Merging Adarsh's work for updating rest api.

Thank you very much Adarsh.

PR #482, opened 2021-01-17
Made causemos integration consume multiple adjectives per edge

manujinda:
if provided. Although input json could provide a list of gradable adjectives, it only provides a single polarity value for all of them. This might be require fixing.

For the moment we are replicating the single polarity value across the whole list of gradable adjectives.


manujinda:
This branch seg faults. Need to fix that before merging.

PR #485, opened 2021-02-15
Delphi year month to epoch

manujinda:
1. Made Delphi work with unix epochs for CauseMos integration portion.
2. Now the delphi local run and CauseMos run are diverged. The CauseMos
    run can only use the continuous (matrix exponential) version to train
    the model. For the moment the local run still uses year month format and
    it can only use to discrete transition matrix format to train the model.
    
    TODO:
    The local run needs to be updated to match the CauseMos run which uses
    the matrix exponential. Afterwards we will not be able to use the
    discretized transition matrix to train the model.


adarshp:
Thank you @manujinda !

PR #488, opened 2021-03-05
Delphi causemos new create model

manujinda:
Updated to use the new CauseMos create model json format.


manujinda:
1. Updated Causemose create model response
Now edge weights are returned as a vector.

2. Adapted the code to circumvent memory problems in Boost Graph Library.
    
    When setS is used as the OutEdgeList in adjacency_list graph
    representation, BGL brings in memory problems when removing a vertex
    (remove_vertex()). Changed it to vecS.
    
    Had to update AnalysisGraph::add_edge() to prevent multiple directed
    edges being added from one vertex o another.
    
    Also added a line to release the random number generator singleton
    object in AnalysisGraph::from_causmos_json_dict() method.

3. Added more test code to create_model.cpp to investigate the segmentation
    fault.
    
    Found  bug in AnalysisGraph::remove_edges() method. Fixed it.

4. Fixed a problem in AnalysisGraph::print_all_paths()

PR #494, opened 2021-09-29
Compiler optimizations

bkj:
This pull request includes two simple compiler optimizations:
- a) Use `-O3 -ffast-math` in `CMakeLists.txt`
- b) Inline the `sqr(x) {return x * x}` function from `lib/utils.cpp`

### Testing
I've also added a build and benchmark script, so we can run w/:
```
./build_and_run.sh
```

### Results
| version     | burnin | posterior sampling |
| ----------- | ------ | ------------------ |
| original code | 9.5s | 1.8s |
| a only        | 3.7s | 0.7s |
| a and b       | 1.2s | 0.2s |

So using both optimizations reduces runtime of this example by ~87%.

### Notes
This is a PR to `master`, but the numbers and experiments above are based on comparison to commit `c07eeae9394ab30ca8d984b2ec2e40ab4c2d2e08`, per @manujinda 's recommendation. _I have not tested on `master`_, but it should be easy for you to copy/paste these changes to whichever branch is currently under development.

### Possible Future Work
After these optimizations, profiling via `callgrind` reveals that the majority of runtime is spent computing calls to `exp` in `KDE::pdf`.  To get additional speedups, we'd need to make `exp` run faster.  This may be possible, depending on how much precision is required, by eg. using approximate `exp` functions like the ones described [here](https://stackoverflow.com/questions/10552280).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,7 +10,7 @@ set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
 set(Boost_NO_BOOST_CMAKE ON)
 set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 set(CMAKE_CXX_FLAGS
-    "${CMAKE_CXX_FLAGS} -fvisibility=hidden -fvisibility-inlines-hidden")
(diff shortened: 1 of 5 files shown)


bkj:
Testing this some more -- it's possible that `-ffast-math` actually breaks something.  If/before this gets merged -- the impact of `-ffast-math` should be tested.  Speedup w/o `-ffast-math` is still substantial (2.9s).

PR #504, opened 2021-11-13
Implemented the CauseMos API version 3.0.0 to a first order

jastier:
This pull request meets the following status for the CauseMos API version 3.0.0:

1. /create-model (DONE)
2. /models/{modelId}/experiments/{experimentId} (DONE)
3. /models/{modelId}/experiments (DONE)
4. /models/{modelId} (DONE)
5. /models/{modelId}/training-progress (DONE)
6. /models/{modelId}/edit-indicators (Endpoint stub with advisory)
7. /models/{modelId}/edit-edges (Endpoint stub with advisory)


manujinda (changes requested):
apps/delphi_rest_api.cpp
Lines: 87 - 145

This is where the projection result is formatted according to the format requested by the API and stored in the DB.

The Causemos API version 3.0.0 specificities a new format and it is the major change in this API. The new format is under the _response_ section of the _/models/{modelId}/experiments/{experimentId}_ section of the new specification.

Without conforming Delphi REST API with this format, we cannot declare victory in implementing the new API.

lib/AnalysisGraph.hpp
```diff
@@ -259,11 +260,17 @@ class AnalysisGraph {
...
+  float training_progress = 0.0;  // Range is [0.0, 1.0]
+
+  // model is completely trained
   bool trained = false;
 
+  // training was stopped by user input
+  bool stopped = false;
+
+  // Used to check whether there is a trained model before calling
+  // generate_prediction()
```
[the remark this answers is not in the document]
jastier: Fixed.

lib/train_model.cpp
```diff
@@ -62,6 +87,16 @@ void AnalysisGraph::run_train_model(int res,
                                 unordered_map<string, double> concept_max_vals,
                                 unordered_map<string, function<double(unsigned int, double)>> ext_concepts
                                 ) {
+
+    cout << "train_model.cpp.run_train_model" << endl;
```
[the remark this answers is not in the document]
jastier: Debug print removed.

jastier:
Fixing the formatting now.

The comment placement and debug output issues are done and on the PR.

jastier:
The experiment data format is fixed.

jastier:
The Projection result data format is fixed.

apps/delphi_rest_api.cpp
```diff
@@ -165,18 +128,11 @@ class Experiment {
...
-                        {"timestamp", timesteps_nparr[i]},
-                        {"value", median_value}};
-
-                    data_dict["values"].push_back(value_dict);
-                    value_dict["value"] = lower_limit;
-                    data_dict["confidenceInterval"]["lower"].push_back(
-                        value_dict);
+		    json tseries;
+		    tseries["timestamp"] = timesteps_nparr[i];
+		    tseries["values"] = vector<double>{lower_limit, median_value,upper_limit};
```
manujinda, line 133: Here we have to send the raw prediction results instead of these. So we just have to send the time_steps
vector.

Looking back at the specification I see the confusion. They have only mentioned that we have to send an array but omitted what the contents of the array. I know this only because I was present at the discussion where this decision was made public by Uncharted.

manujinda (changes requested):

apps/delphi_rest_api.cpp
jastier: OK, the entire time_step vector is now sent for each timestamp.

apps/delphi_rest_api.cpp
```diff
@@ -132,51 +95,32 @@ class Experiment {
...
                 json data_dict;
                 data_dict["concept"] = conceptname;
-                data_dict["values"] = vector<unordered_map<string, double>>{};
-                unordered_map<string, vector<double>> m{
-                    {"upper", vector<double>{}}, {"lower", vector<double>{}}};
-                data_dict["confidenceInterval"] = m;
+                data_dict["values"] = vector<json>{};
                 for (int i = 0; i < timestamp_sample_matrix.size(); i++) {
                     vector<double> time_step = timestamp_sample_matrix[i];
                     sort(time_step.begin(), time_step.end());
```
manujinda, line 112: We could remove this sort to improve speed.

apps/delphi_rest_api.cpp
```diff
@@ -132,51 +95,32 @@ class Experiment {
                 about-statistical-courses/research-methods-and-statistics/chapter-8-content-8
             */
             int n = G.get_res();
-
             int lower_rank = (int)((n - 1.96 * sqrt(n)) / 2);
             int upper_rank = (int)((2 + n + 1.96 * sqrt(n)) / 2);
-
             lower_rank = lower_rank < 0 ? 0 : lower_rank;
             upper_rank = upper_rank >= n ? n - 1 : upper_rank;
```
manujinda, line 101: We could remove rank related code.

apps/delphi_rest_api.cpp
```diff
@@ -132,51 +95,32 @@ class Experiment {
...
-                data_dict["values"] = vector<unordered_map<string, double>>{};
-                unordered_map<string, vector<double>> m{
-                    {"upper", vector<double>{}}, {"lower", vector<double>{}}};
-                data_dict["confidenceInterval"] = m;
+                data_dict["values"] = vector<json>{};
                 for (int i = 0; i < timestamp_sample_matrix.size(); i++) {
                     vector<double> time_step = timestamp_sample_matrix[i];
                     sort(time_step.begin(), time_step.end());
                     int l = time_step.size() / 2;
                     double median_value = 0;
```
manujinda, line 114: We don't need the median

apps/delphi_rest_api.cpp
```diff
@@ -132,51 +95,32 @@ class Experiment {
...
-                    if (time_step.size())
-                        median_value =
-                            time_step.size() % 2
-                                ? time_step[l]
-                                : (time_step[l] + time_step[l - 1]) / 2;
-
+                    if (time_step.size()) median_value = time_step.size() % 2 ? 
+                        time_step[l] : (time_step[l] + time_step[l - 1]) / 2;
                     double lower_limit = time_step[lower_rank];
                     double upper_limit = time_step[upper_rank];
```
manujinda, line 118: We don't need to find median, lower and upper limits

manujinda (approved):
Looks according to the new CauseMos API. 

Added few comments where we could gain some speed by removing unused calculations.

We could go ahead merge and let Uncharted know.

PR #509, opened 2021-11-23
Dockerfile support for delphi_rest_api

jastier:
The ~/delphi/Dockerfile will now build and run the delphi_rest_api application.
CauseMos API version 3.0.0 is implemented.


adarshp (changes requested):
@jastier - the Dockerfile doesn't need to do anything Python-related since we don't expect to use it for Jupyter notebooks, etc. So I think you can get rid of lines 21-22, 28, 46, and 48.

adarshp (approved):

PR #513, opened 2021-11-24
Reduced the sampling resolution to 100 in all cases

jastier:
The sampling resolution is now set to 100 in all run modes.


manujinda (approved):
looks good

PR #537, opened 2022-02-07
edit-edges fixes

jastier:
- The edit-edges endpoint will now check that the model exists and has finished training before freezing edges.
- The return value from the AnalysisGraph::freeze_edge_weight is now read.
- Cleaned up the state control when models are training.


jastier:
Removing this PR because for some reason the CI is taking too long.

PR #549, opened 2022-02-14
Busy status joseph

jastier:


jastier:
## Better control of Model busy states outside of training

- create-model endpoint will not overwrite a busy model
- edge edges endpoints will freeze edges on a busy model
- experiment create endpoint will not use a busy model

